PR #265, opened 2014-12-24
Auto-playlist and lookup improvements

Eiron:
array_merge can have some funny behaviour from what I have been reading. For instance if $track has key [0] and $similar allocates keys starting at [0], then it may overwrite the original track. This may fix that. If not, swapping the lines (as they were before) will return it to its original behaviour, albeit with the similar tracks appearing above the original track.

I also changed the 100 in my group chart thing to sizeof($keys), as this is a bit neater and more explicit.

If this does not fix the missing $track issue, then I will request a dlog to check what the array_unique is working with.


Flump5281:
Some of these changes have been implemented in #290. Others are silly, not longer relevant or need updating to reflect the changes this makes.

Eiron:
Content with what has been done. Happy to close this now.

PR #280, opened 2014-12-29
Bootstrap

Flump5281:
I've tried to keep this as client side as possible, but unfortunately that didn't work - in some cases server-side changes were necessary, in others I may have forgotten...

Changes in a nutshell:
- Removes some more leftover version broker code
- Replaces the general page template layout with Bootstrap
- Replaces jQueryUI icons with Bootstrap icons
- Replaces jQueryUI state-highlight and helper-hiddens with Bootstrap alerts and hiddens
- Replaces jQueryUI Slider on Show Planner with a home-grown slider
-- Adds support for viewing the intro on a track
-- Adds support for setting a cue point on a track
- Replaces jQueryUI datetimepicker plugin with Bootstrap-based picker
- Replaces jQueryUI autocompletes with typeahead.js (and Bloodhound)
-- Lots of potential to do more cool stuff here later
- Replaces jQueryUI dialogs with myury.createDialog function
-- Ajax requests now serve a 401 when a login is required
-- Global ajax handler now responds to 401s by redirecting the user to the login page
- Replaces jQuery contextmenu plugin with home-made context menu that uses Bootstrap
-- Fixes Show Planner Automatic Advance/Play On Load functionality
-- Adds support for the Repeat options in Show Planner
-- Adds support for deleting items using the context menu in Show Planner
- Replaces jQueryUI's js file with one that just contains the Sortable plugin (now much lighter!)
- Moves datatables away from jQueryUI plugin, uses Bootstrap styles instead
- General polish to the Show Planner upload functionality
- Adds Moment.js (currently only loaded in forms)
- Fixes some requests being POSTs where GETs are sufficient
- Replaces the Schedule page iFrame to ury.org.uk/schedule with a simple schedule renderer
- Fixes and re-enables schedule conflict detection
- Fixes some issues with the metadata system
- Complete rewrite of SIS client-side functionality to use Bootstrap and a better plugin system

Draft deploy plan:
- Make /myradio point to /usr/local/www/myradio instead of myradio-vproxy
- Make /myury a 301 permanent redirect to /myradio
- Update /usr/local/www/myradio-dev to have a baseurl of /myradio-dev again
- Pull the changes in ury:/usr/local/www/myradio and uryfs1:/var/www/myradio


.gitignore
```diff
@@ -5,4 +5,6 @@ util/checkpassword.conf
 *.sw[a-z]
 sftp-*.json
 .tags*
-.phpintel
\ No newline at end of file
+<<<<<<< HEAD
```
LordAro: lolmerge

src/Classes/NIPSWeb/NIPSWeb_ManagedItem.php
```diff
@@ -214,15 +214,22 @@ public static function storeItem($tmpid, $title)
         }
         $dbfile = $item->getFolder().'/'.$item->getID();
 
-        //Convert it with ffmpeg
+        //Convert it with ffmpeg/avconv
+        if (`which ffmpeg`) {
```
LordAro: This... works? Ew? Is there no nicer/better way of doing so?

Flump5281: Eww indeed. Feel free to look at the various ways of calling things in PHP to investigate. This is more-or-less duplication of what's already in Track.

src/Public/js/myradio.schedule.view.js
```diff
@@ -0,0 +1,273 @@
...
+
+	//Set up view
+	scheduleContainerDiv.style.position = 'relative';
+	scheduleContainerDiv.className = 'scheduleview scheduleview-view-' + view;
+	options.container.appendChild(scheduleContainerDiv);
+	loadWeek();
+
+	return {
+
+	};
```
LordAro: ?

schema/base.sql
```diff
@@ -7270,3 +7262,38 @@ CREATE TABLE music.explicit_checked (
 
 ALTER TABLE ONLY music.explicit_checked ADD CONSTRAINT explicit_checked_pkey PRIMARY KEY (trackid);
 ALTER TABLE music.explicit_checked ADD CONSTRAINT explicit_checked_fkey FOREIGN KEY (trackid) REFERENCES public.rec_track(trackid) ON DELETE CASCADE;
+
+SET search_path = myury, pg_catalog;
```
AJDurant: debranding?

Flump5281: Moving the myury. tables to myradio. is probably a separate piece of work..

src/Classes/ServiceAPI/MyRadio_Webcam.php
```diff
@@ -69,32 +69,39 @@ public static function getArchiveTimeRange()
      */
     public static function getCurrentWebcam()
     {
-        $current = file_get_contents(Config::$webcam_current_url);
+        if (Config::$webcam_current_url) {
+            $current = (int)file_get_contents(Config::$webcam_current_url);
```
AJDurant: Will this cause an exception/error if the file actually contains a string (ie a URL) as that is what the default fall-through is there for - yes it's a horrible hack, and the form to allow setting a URL is not available to users.

Flump5281: It won't throw an exception - in this case the value will be 0 (see http://php.net/manual/en/language.types.string.php#language.types.string.conversion)

src/Controllers/Setup/checks.php
```diff
@@ -183,79 +183,92 @@ function convertPHPSizeToBytes($sSize)
...
-                  }
-              } else {
-                  echo '<h3>The following tests failed, but they aren\'t required for MyRadio to run:</h3><ul>';
-                  foreach ($warnings as $warning) {
-                      echo '<li>'.$warning.'</li>';
-                  }
-                  echo '</ul>';
-              }
+    <script type="text/javascript" src="js/vendor/jquery-2.1.0.min.js"></script>
+    <script type="text/javascript" src="js/vendor/bootstrap.min.js"></script>
```
AJDurant: Loading scripts in the header? You render blocker

Flump5281: This is in setup so isn't the end of the world, and is how it did it before (although I do agree that should be moved) - not sure whether it's worth doing in this or in the later general JS refactor I have planned.

src/Public/js/myradio.schedule.view.js
```diff
@@ -0,0 +1,273 @@
...
+					addNowMarker(dayDiv);
+					nextFound = true;
+				}
+				if (view === 'week') {
+					showDiv.style.height = height;
+				} else {
+					showDiv.style.minHeight = height;
+				}
+				dayDiv.appendChild(showDiv);
+				lastTime = endTime.unix();
```
AJDurant: This will still break stuff, remove the .unix() on endTime

src/Public/js/myury.core.js
```diff
@@ -9,46 +9,74 @@ window.myury = {
...
-                    });
-                }
-            }
+        myury.createDialog(
+            'Error',
+            '<p>It looks like that request worked, but things might not quite work as expected.</p><details>' + myradio_errors + '</details>',
+            [myury.closeButton(), myury.reportButton(xhr, settings, myradio_errors)]
+        );
+    },
+    createDialog: function(title, text, buttons, startHidden) {
```
AJDurant: Isn't their bootstrap js for this?

Flump5281: http://getbootstrap.com/javascript/#modals
Nope.

src/Public/js/myury.form.js
```diff
@@ -311,6 +375,7 @@ window.MyRadioForm = {
      */
     if ($('#UPLOAD_IDENTIFIER').length !== 0) {
       $('form').on('submit', function() {
+        console.log('UPLOAD HANDLER');
```
AJDurant: needed?

src/Public/js/nipsweb.librarypane.js
```diff
@@ -58,12 +59,14 @@ $(document).ready(function() {
...
         url: myury.makeURL('MyRadio', 'a-findtrack'),
-        type: 'post',
+        type: 'get',
         data: {itonesplaylistid: $(this).val().replace(/managed-/, ''), digitised: true, limit: 0},
         success: function(data) {
           for (file in data) {
             $('#baps-channel-res').append(
                     '<li id="' + data[file].album.recordid + '-' + data[file].trackid +
+                    '" title="' + data[file].title + '(' + data[file].length + ')' +
+                    '" intro="' + data[file].intro + '"' +
```
AJDurant: this kind of thing should probably be a data-* attribute

src/Public/js/nipsweb.player.js
```diff
@@ -1,685 +1,1105 @@
...
+
+        ajaxQueue.queue(function(next) {
+            $('#notice').html('Saving changes...').show();
+            $.ajax({
+                async: !addOp,
+                cache: false,
+                success: function(data) {
+                    $('#notice').hide();
+                    for (i in data) {
+                        if (i === 'myury_errors') {
```
AJDurant: myury_errors or myradio_errors, I think this is now sent as the latter

src/Public/js/nipsweb.player.js
```diff
@@ -1,685 +1,1105 @@
...
+    */
+    var shipChanges = function(ops, addOp, pNext) {
+        if (typeof addOp === 'undefined') {
+            addOp = false;
+        }
+
+        ajaxQueue.queue(function(next) {
+            $('#notice').html('Saving changes...').show();
+            $.ajax({
+                async: !addOp,
```
AJDurant: Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience. For more help, check http://xhr.spec.whatwg.org/.

src/Templates/NIPSWeb/live.twig
```diff
@@ -23,7 +23,7 @@
             <p>Connecting to server...</p>
             <div id="init-progressbar"></div>
         </div>
-        <div class="ui-dialog-buttonpane ui-widget-content ui-helper-hidden" id="init-overlay-fallback">
+        <div class="ui-dialog-buttonpane ui-widget-content hidden" id="init-overlay-fallback">
```
AJDurant: this file hasn't been updated and doesn't match main.twig for show planner - tho I'm guessing it doesn't work atm anyway

src/Templates/NIPSWeb/main.twig
```diff
@@ -4,39 +4,38 @@
...
   <h2>{{title}}</h2>
 </header>
 <div id="grid" class="transBG clearfix">
-
-  <div id="bapswrapper" class="ui-widget">
+  <div id="bapswrapper">
     <div id="baps-channel-container" class="clearfix">
       {% for i in 0..2 %}
-        <ul class="box width-1 height-2 baps-channel ui-widget-content" channel="{{i+1}}" id="baps-channel-{{i + 1}}">
+        <ul class="box col-md-3 col-sm-3 col-xs-3 height-2 baps-channel" channel="{{i+1}}" id="baps-channel-{{i + 1}}">
```
AJDurant: pretty sure you only need to specify the smallest one if they are all the same

src/Templates/NIPSWeb/manage_library.twig
```diff
@@ -3,24 +3,23 @@
...
               <a href="http://www.last.fm/">Powered by <img src="http://cdn.last.fm/flatness/badges/lastfm_black_small.gif" /></a>
-          </div></div>
-          <div class="ui-state-highlight" id="central-status"></div>
-          <div class="ui-widget-content" id="central-result"></div>
+          </div><div class="alert alert-warning" id="central-status">Getting set up...</div></div>
+          <div id="central-result"></div>
         </div>
-        <div id="managed-container" class="res-container ui-widget-content ui-helper-hidden">Managed playlists are controlled using <a href="/itones/" target="_blank">Campus Playout Manager</a>. Entering them there also adds them to the wonderful Campus Jukebox</div>
-        <div id="res-container" class="res-container ui-widget-content ui-helper-hidden">
+        <div id="managed-container" class="res-containert hidden">Managed playlists are controlled using <a href="/itones/" target="_blank">Campus Playout Manager</a>. Entering them there also adds them to the wonderful Campus Jukebox</div>
```
AJDurant: typo

src/Templates/SIS/tabs/help.twig
```diff
@@ -1,8 +1,8 @@
 <div id="gettingstarted">
   <button id="hide-help">Don't show me this in future</button>
-  <h2>Welcome to SIS!</h2>
+  <h2>Welcome the Studio Information Service!</h2>
```
AJDurant: help has been removed, this is a dead template

src/Templates/form.twig
```diff
@@ -25,18 +26,20 @@
         {% if captcha %}
         <div class="captcha">{{ captcha | raw }}</div>
         {% endif %}
-        <input type="submit" id="{{frm_name}}-submit" value="Submit" />
+        <input type="submit" class="btn btn-primary" id="{{frm_name}}-submit" value="Submit" />
     </form>
     <div class="myradiofrm-file-upload-progress"><div class="progress-label">Starting Upload...</div></div>
-    <iframe class="ui-helper-hidden" id="myradiofrm-file-upload-iframe"></iframe>
+    <iframe class="hidden hidden" id="myradiofrm-file-upload-iframe"></iframe>
```
AJDurant: double class

src/Templates/parts/table-footer.twig
```diff
@@ -1,8 +1,8 @@
 <script type="text/javascript" src="{{baseurl}}js/vendor/jquery.dataTables.min.js"></script>
+<script type="text/javascript" src="{{baseurl}}js/vendor/dataTables.bootstrap.js"></script>
 <script type="text/javascript">
   //Set DataTable defaults
   $.extend($.fn.dataTable.defaults, {
-    bJQueryUI: true,
```
AJDurant: why was this moved from defaults into every file?

src/Public/js/nipsweb.managelibrary.js
```diff
@@ -1,258 +1,304 @@
...
+                    }));
+
+                    $('#res-result').append(result);
+                }
+            }
+        });
+    };
+
+    var initialise = function() {
+        $('#res-type-sel').on('change', res_type_sel_change_handler);
```
AJDurant: this is probably the only event needed to track.

src/Public/js/nipsweb.player.js
Flump5281: Just checked, still myury_errors in all uses I can see.

src/Public/js/nipsweb.player.js
Flump5281: I know - that's not a new thing and I did try it without but I think it needs testing separately to everything else (we historically had issues with changes being processed in the wrong order)

src/Templates/NIPSWeb/live.twig
Flump5281: There's a low-priority Issue for that - live is not a production thing and likely doesn't work (hasn't even been renamed for Rapid yet!)

src/Templates/NIPSWeb/main.twig
Flump5281: Ah, I added sm and xs because just md made smaller screens flatten them to full width.

src/Templates/parts/table-footer.twig
Flump5281: It hasn't been. A long time ago it was in every file and removed. Now it's not using bJQueryUI - it's bootstrap now!

src/Public/js/nipsweb.managelibrary.js
Flump5281: change in chrome doesn't fire until the element has been blur'd sometimes - they're needed.

src/Templates/NIPSWeb/live.twig
wlcx: *Bifrost, jeez get it right

AJDurant: The point that @Flump5281 was making is that it is still using the name behind the previous name...

PR #310, opened 2015-01-12
Fixes updateCacheObject not working

Flump5281:


Flump5281:
Tested and confirmed working on dev.

PR #315, opened 2015-01-13
Tidy up scheduler show listing

wlcx:
Don't merge yet, opening for comments


src/Controllers/Scheduler/allShows.php
```diff
@@ -11,7 +11,7 @@
 use \MyRadio\ServiceAPI\ServiceAPI;
 use \MyRadio\ServiceAPI\MyRadio_Show;
 
-$shows = MyRadio_Show::getAllShows();
+$shows = MyRadio_Show::getAllShows(1, true);
```
AJDurant: looking at this function call from here you wouldn't know what 1 and true mean, so it should have a comment explaining

src/Controllers/Scheduler/allShows.php
```diff
@@ -11,7 +11,7 @@
 use \MyRadio\ServiceAPI\ServiceAPI;
 use \MyRadio\ServiceAPI\MyRadio_Show;
 
-$shows = MyRadio_Show::getAllShows();
+$shows = MyRadio_Show::getAllShows(1, true); # show_type_id=1, current_term_only=true
```
AJDurant: use // and put it on the line above for clarity. Also it is clearer to say something in english rather than computer eg:
Get public shows (type 1) from this term only (true)

src/Controllers/Scheduler/shows.php
```diff
@@ -11,7 +11,8 @@
 use \MyRadio\ServiceAPI\ServiceAPI;
 use \MyRadio\ServiceAPI\MyRadio_Show;
 
-$shows = MyRadio_Show::getAllShows();
+// Get public shows (type 1) and check if all=true, in which case all shows, not just this term's (false)
+$shows = MyRadio_Show::getAllShows(1, !( isset($_GET["all"]) && ($_GET["all"] == 'true')));
```
AJDurant: it's normally clearer to do this kind of thing into a variable which is then fed into the function call

AJDurant:
otherwise :+1:

src/Classes/ServiceAPI/MyRadio_Show.php
```diff
@@ -654,20 +654,42 @@ public function setCredits($users, $credittypes, $table = null, $pkey = null)
...
-                SELECT metadata_value FROM schedule.show_metadata
-                WHERE show.show_id=show_metadata.show_id AND metadata_key_id=2
-                AND effective_from <= NOW()
-                AND (effective_to IS NULL OR effective_to > NOW())
-                ORDER BY effective_from DESC LIMIT 1
-            );',
-            [$show_type_id]
-        );
+        if($current_term_only)
+        {
```
Flump5281: Should be on same line as if

src/Classes/ServiceAPI/MyRadio_Show.php
```diff
@@ -654,20 +654,42 @@ public function setCredits($users, $credittypes, $table = null, $pkey = null)
...
+            $show_ids = self::$db->fetchColumn(
+                'SELECT show_id FROM schedule.show
+                WHERE show_type_id=$1 AND EXISTS (
+                        SELECT * FROM schedule.show_season
+                        WHERE schedule.show_season.show_id=schedule.show.show_id
+                        AND schedule.show_season.termid=$2
```
Flump5281: May be worth considering whether you want to include/exclude seasons that don't actually have any scheduled timeslots

src/Classes/ServiceAPI/MyRadio_Show.php
```diff
@@ -654,20 +654,42 @@ public function setCredits($users, $credittypes, $table = null, $pkey = null)
...
+                )
+                ORDER BY (
+                    SELECT metadata_value FROM schedule.show_metadata
+                    WHERE show.show_id=show_metadata.show_id AND metadata_key_id=2
```
Flump5281: Is that metadata_key_id harcoded elsewhere too? I'm guessing yes...

Flump5281: Just answered my own question - probably an enhancement to raise for another time

src/Controllers/Scheduler/shows.php
Flump5281: Also, === where possible, and I prefer $_REQUEST

src/Classes/ServiceAPI/MyRadio_Show.php
wlcx: I'd argue that because it's "internal" these probably should be included.

Flump5281: I didn't have an opinion on it, was just checking you knew it'd return empty ones too :)

:+1:

src/Classes/ServiceAPI/MyRadio_Show.php
AJDurant: There should also be a space between the if and the ( ;)

wlcx: Submit a pull request :p

PR #319, opened 2015-01-15
Coreutils

LordAro:
Might be some more stuff that could be done, say if you think so

### DIFF
--- a/src/Classes/Database.php
+++ b/src/Classes/Database.php
@@ -6,7 +6,7 @@
 
 namespace MyRadio;
 
-use \MyRadio\MyRadio\CoreUtils;
+use \MyRadio\MyRadio\AuthUtils;
 
 /**
  * This singleton class handles actual database connection
@@ -105,7 +105,7 @@ class Database
             }
         }
 
-        if (isset($_REQUEST['dbdbg']) && CoreUtils::hasPermission(AUTH_SHOWERRORS)) {
+        if (isset($_REQUEST['dbdbg']) && AuthUtils::hasPermission(AUTH_SHOWERRORS)) {
             //Debug output
             echo $sql . '&nbsp;' . print_r($params, true) . '<br>';
         }
--- a/src/Classes/MyRadio/MyRadioDefaultAuthenticator.php
+++ b/src/Classes/MyRadio/MyRadioDefaultAuthenticator.php
@@ -7,6 +7,8 @@ use \MyRadio\Database;
 use \MyRadio\MyRadioEmail;
 use \MyRadio\MyRadioException;
 use \MyRadio\ServiceAPI\MyRadio_User;
+use \MyRadio\MyRadio\CoreUtils;
+use \MyRadio\MyRadio\URLUtils;
 
 /**
  * An Authenticator processes login requests for a user against a specific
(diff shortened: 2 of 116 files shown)


src/Classes/Database.php
```diff
@@ -105,7 +105,7 @@ public function query($sql, $params = [], $rollback = false)
             }
         }
 
-        if (isset($_REQUEST['dbdbg']) && CoreUtils::hasPermission(AUTH_SHOWERRORS)) {
+        if (isset($_REQUEST['dbdbg']) && AuthUtils::hasPermission(AUTH_SHOWERRORS)) {
```
Flump5281: Heh, actually I want to deprecate hasPermission to explicitly use the User class at some point, but that's good for now.

src/Classes/MyRadio/MyRadioDefaultAuthenticator.php
```diff
@@ -7,6 +7,8 @@
 use \MyRadio\MyRadioEmail;
 use \MyRadio\MyRadioException;
 use \MyRadio\ServiceAPI\MyRadio_User;
+use \MyRadio\MyRadio\CoreUtils;
```
Flump5281: Well that explains a lot.

Flump5281: No wait - both these are unneccessary - this file is in the same namespace.

src/Classes/MyRadio/URLUtils.php
```diff
@@ -0,0 +1,140 @@
...
+    public static function nocontent()
+    {
+        header('HTTP/1.1 204 No Content');
+        exit;
+    }
+
+    /**
+     * Responds with JSON data.
+     */
+    public static function dataToJSON($data)
```
Flump5281: Is this a "URL" util? /me not sure

Flump5281:
First read through I don't see anything major, but have a few points or questions:

Is it worth moving ***Utils into its own namespace?
How easy is this change going to apply to the bootstrap branch?
Is it worth basing this on the bootstrap branch and doing it at the same time, or perhaps leaving this open until bootstrap is ready to go in then start testing this?

AJDurant:
I'd say wait till after bootstrap

LordAro:
depends how much longer bootstrap is going to take, i guess
dataToJSON - i moved it because it contained header() functions so i figured it was URL-related "enough"
i'll look at the other issues - even the hasPermission thing, if i'm bored enough

Flump5281:
Bootstrap has just the tracklisting tab and user testing left - shouldn't be long at all now.

src/Classes/MyRadioException.php
```diff
@@ -94,9 +95,9 @@ public function uncaught()
             //Configuration is available, use this to decide what to do
             if (!$silent
                 && Config::$display_errors
-                || (class_exists('\MyRadio\MyRadio\CoreUtils')
+                || (class_exists('\MyRadio\MyRadio\AuthUtils')
```
Flump5281: I'm not sure how class_exists behaves - does it check if it can be loaded via autoloader?
CoreUtils could be checked directly as it was needed for early in the setup process, although AuthUtils might be too.

Needs checking, I think.

Flump5281:
Can you merge with latest changes? Then I'm happy to go with this.

PR #320, opened 2015-01-16
Update Demo labeling to Studio Training

AJDurant:


wlcx:
Is 'demo' being kept as internal parlance for module names/urls? Worth changing them too?

Flump5281:
Not worth the effort changing configs and database things for that.

Besides, this is one of the features I want to remove anyway, as it's actually entirely a hack.

AJDurant:
Yeah, this was just a quick re-nameing thing, at some point it will be completely re-worked, but for now it is just frontend

Flump5281:
Should possibly consistently use Session or Slot?

AJDurant:
I wondered that, but was mostly happy with the wording

PR #324, opened 2015-01-27
Add ability to sign Contract via MyRadio

LordAro:
Requires adding a 'contract_signed' column to members

### DIFF
--- a/src/Classes/Config.php
+++ b/src/Classes/Config.php
@@ -241,6 +241,12 @@ final class Config
     public static $vacant_officer_uri = '/media/image_meta/MyRadioImageMetadata/32.jpeg';
 
     /**
+     * The full web address of a copy of the Presenters Contract.
+     * @var String
+     */
+    public static $contract_uri = '/wiki/Contract';
+
+    /**
      * The file system path to the Central Database. Must be absolute. Can not be smb://, but may be a network share
      * mounted to the file system mountpoint.
      * @var String
@@ -287,7 +293,7 @@ final class Config
      * @var String
      */
     public static $lastfm_api_secret;
-    
+
     /**
      * The last.fm group specifically for the University of York. If using
      * this aspect of the code you probably want to change this bit.
--- a/src/Classes/Config.php
+++ b/src/Classes/Config.php
@@ -242,6 +242,7 @@ final class Config
 
     /**
      * The full web address of a copy of the Presenters Contract.
+     * @note If this variable is empty, then contracts are disabled.
      * @var String
      */
     public static $contract_uri = '/wiki/Contract';
--- a/src/Classes/Config.php
+++ b/src/Classes/Config.php
@@ -245,7 +245,7 @@ final class Config
      * @note If this variable is empty, then contracts are disabled.
      * @var String
      */
-    public static $contract_uri = '/wiki/Contract';
+    public static $contract_uri = '';
 
     /**
      * The file system path to the Central Database. Must be absolute. Can not be smb://, but may be a network share
--- a/src/Classes/ServiceAPI/MyRadio_User.php
+++ b/src/Classes/ServiceAPI/MyRadio_User.php
@@ -200,7 +200,7 @@ class MyRadio_User extends ServiceAPI
             'SELECT fname, sname, sex, college AS collegeid, l_college.descr AS college,
             phone, email, receive_email, local_name, local_alias, eduroam,
             account_locked, last_login, joined, profile_photo, bio,
-            auth_provider, require_password_change
+            auth_provider, require_password_change, contract_signed
             FROM member, l_college
             WHERE memberid=$1
             AND member.college = l_college.collegeid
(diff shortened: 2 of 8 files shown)


src/Classes/ServiceAPI/MyRadio_User.php
```diff
@@ -1372,6 +1401,17 @@ public function getEditForm()
                         ['value' => 'o', 'text' => 'Other']
                     ]
                 ]))
+                ->addField(new MyRadioFormField('contract', MyRadioFormField::TYPE_CHECK, [
```
Flump5281: Isn't this supposed to be an option only checkable by someone PC-y (i.e. they sign a physical piece of paper)?

src/Controllers/MyRadio/default.php
```diff
@@ -32,7 +32,8 @@
         ->addVariable('studio_demoed', $user->isStudioDemoed())
         ->addVariable('is_trainer', $user->isTrainer())
         ->addVariable('has_show', $user->hasShow())
-        ->addVariable('paid', $user->isCurrentlyPaid());
+        ->addVariable('paid', $user->isCurrentlyPaid())
+        ->addVariable('contract_signed', $user->hasSignedContract());
```
Flump5281: Starting to think we should just expose $user->toDataSource generally... but that's a seperate issue

src/Classes/ServiceAPI/MyRadio_User.php
LordAro: no, the point was to save paper by letting them do it here
(coincidentally, we couldn't find any paper for the printer at the time)

Flump5281:
- [ ] Make this actually have an effect on the ability to use certain functionality
- [ ] Add the ability to disable this feature in the Config file

AJDurant:
[ ] Contract up to date in wiki
[ ] Wording checked with PC
[ ] column added to db

wlcx:
Column signed_contract of type boolean defaulting to false has been added to database.

src/Classes/ServiceAPI/MyRadio_User.php
```diff
@@ -1371,8 +1406,23 @@ public function getEditForm()
                         ['value' => 'f', 'text' => 'Female'],
                         ['value' => 'o', 'text' => 'Other']
                     ]
-                ]))
-                ->addField(new MyRadioFormField('sec_personal_close', MyRadioFormField::TYPE_SECTION_CLOSE));
+                ]));
+        if (empty(Config::$contract_uri) === false) {
+            $form->addField(new MyRadioFormField('contract',
+                empty(Config::$contract_uri) ? MyRadioFormField::TYPE_HIDDEN : MyRadioFormField::TYPE_CHECK, [
```
AJDurant: this check has already been made

AJDurant:
Still doesn't affect ability to do stuff - ie apply for shows.
Also if contracts are disabled (the url isn't set) then it should not block applying for shows.
I would suggest that if it is disabled then the db value defaults to true, so the check is always completed, and you don't have to duplicate the disabled check.

src/Classes/Config.php
```diff
@@ -241,6 +241,13 @@
     public static $vacant_officer_uri = '/media/image_meta/MyRadioImageMetadata/32.jpeg';
 
     /**
+     * The full web address of a copy of the Presenters Contract.
+     * @note If this variable is empty, then contracts are disabled.
+     * @var String
+     */
+    public static $contract_uri = '/wiki/Contract';
```
Flump5281: Can we default this to null please?

src/Classes/ServiceAPI/MyRadio_User.php
```diff
@@ -629,6 +636,20 @@ public function getAuthProvider()
...
+     * Get whether the user has signed the Presenter Contract
+     *
+     * @return bool if the presenter has signed the contract
+     */
+    public function hasSignedContract()
+    {
+        if (empty(Config::$contract_uri)) {
+            return true;
+        } else {
+            return $this->contract_signed;
```
Flump5281: I don't see any change to the constructor to pull this value from the database, so it's undefined?

src/Classes/ServiceAPI/MyRadio_User.php
Flump5281: :+1:

src/Classes/ServiceAPI/MyRadio_User.php
```diff
@@ -1567,7 +1617,8 @@ public static function create(
                 $email,
                 $receive_email,
                 $eduroam,
-                true
+                true,
+                false
```
Flump5281: Since it's default false, this isn't necessary

src/Controllers/Profile/edit.php
```diff
@@ -35,7 +35,8 @@
         ->setEmail($data['email'])
         ->setReceiveEmail($data['receive_email'])
         ->setEduroam($data['eduroam'])
-        ->setBio($data['bio']);
+        ->setBio($data['bio'])
+        ->setContractSigned($data['contract']);
```
Flump5281: Since there's situations where this value isn't in the form, this'll throw a notice and try and set it to null

src/Classes/ServiceAPI/MyRadio_User.php
```diff
@@ -1371,8 +1406,22 @@ public function getEditForm()
                         ['value' => 'f', 'text' => 'Female'],
                         ['value' => 'o', 'text' => 'Other']
                     ]
-                ]))
-                ->addField(new MyRadioFormField('sec_personal_close', MyRadioFormField::TYPE_SECTION_CLOSE));
+                ]));
+        if (empty(Config::$contract_uri) === false) {
```
Flump5281: if (!empty), although thinking about it, that's personal preference, so ignore me.

PR #326, opened 2015-01-28
Add explicit button

adfw:
Not ready for merging yet, pending discussion regarding checkbox or radio buttons, but might as well be available for code review and CI.


src/Classes/ServiceAPI/MyRadio_Track.php
```diff
@@ -609,6 +609,13 @@ public static function identifyAndStoreTrack($tmpid, $title, $artist, $album, $p
             $ainfo['duration'] = intval($getID3->analyze(Config::$audio_upload_tmp_dir . '/' . $tmpid)['playtime_seconds']);
         }
 
+        // See if the explicit checkbox was set, and set the value for the DB accordingly
+        if (isset($explicit)){
```
Flump5281: Nonononononononono :P

src/Classes/ServiceAPI/MyRadio_Track.php
```diff
@@ -619,7 +626,8 @@ public static function identifyAndStoreTrack($tmpid, $title, $artist, $album, $p
                         'digitised' => true,
                         'duration' => $ainfo['duration'],
                         'recordid' => $ainfo['album']->getID(),
-                        'number' => $ainfo['position']
+                        'number' => $ainfo['position'],
+                        'clean' => $explicit
```
AJDurant: clean = explicit?

src/Classes/ServiceAPI/MyRadio_Track.php
```diff
@@ -586,12 +586,13 @@ public static function identifyUploadedTrack($path)
         }
     }
 
-    public static function identifyAndStoreTrack($tmpid, $title, $artist, $album, $position)
+    public static function identifyAndStoreTrack($tmpid, $title, $artist, $album, $position, $explicit)
```
Flump5281: I think there's still some thinking to be done on this option.

I'm currently thinking it has three values, one of which is default (making the parameter optional)
null (default): It is unknown if this track is explicit
true: Yes, it is explicit
false: No, it isn't explicit

confirm_central_upload can work with that, and then future things using identifyAndStoreTrack will hopefully not have ambiguity.

src/Controllers/NIPSWeb/confirm_central_upload.php
```diff
@@ -15,7 +15,8 @@
     $_REQUEST['title'],
     $_REQUEST['artist'],
     $_REQUEST['album'],
-    $_REQUEST['position']
+    $_REQUEST['position'],
+    $_REQUEST['explicit']
```
AJDurant: something like:
isset(request) ? request === true : null;
would be my suggestion

src/Classes/ServiceAPI/MyRadio_Track.php
AJDurant: this seems logical :+1:

AJDurant:
Will need to inform Jay when this is live, for him to contact all Uploader trained people

PR #382, opened 2015-02-24
Add location to SIS messages

Flump5281:


AJDurant:
Ignores location[1]

Flump5281:
Does not.

AJDurant:
You've done a file permission change 100644 → 100755

PR #387, opened 2015-03-01
Add factories for show

Flump5281:
Substantially improves startup time and database load for fetching all shows on an empty cache. Provides the ServiceAPI functionality to expand this further in future.

- Update image_metadata type to TEXT
- Deprecate Database::decodeArray
- Add default factory method to serviceapi
- Re-write show select query to support returning multiple rows
- Replace getAllShows iterator with single query
- Expand CacheProvider interface with getAll feature

### DIFF
--- a/src/Interfaces/CacheProvider.php
+++ b/src/Interfaces/CacheProvider.php
@@ -23,7 +23,7 @@ interface CacheProvider extends Singleton
      */
     public function get($key);
     /**
-     * Gets all cache entries starting with the given prefic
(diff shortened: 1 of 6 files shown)


src/Interfaces/CacheProvider.php
```diff
@@ -23,6 +23,11 @@ public function set($key, $value, $expires = 0);
      */
     public function get($key);
     /**
+     * Gets all cache entries starting with the given prefic
```
AJDurant: prefic

PR #389, opened 2015-03-03
Access Control and OPTIONS

Flump5281:
Not the best branch name, whoops.

### DIFF
--- a/src/PublicAPI/index.php
+++ b/src/PublicAPI/index.php
@@ -35,6 +35,7 @@ function api_error($code, $message = null, $previous = null)
         401 => "Unauthorized",
         403 => "Forbidden",
         404 => "File Not Found",
+        405 => "Method Not Allowed",
         500 => "Internal Server Error"
     ];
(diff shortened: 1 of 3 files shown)


src/PublicAPI/index.php
```diff
@@ -159,6 +160,31 @@ function invokeArgsNamed(ReflectionMethod $refmethod, $object, Array $args = [])
...
 
 /**
+ * Check and provide Access-Control-Allow-Origin if needed
+ */
+if (isset($_SERVER['HTTP_ORIGIN'])) {
+    if (
+        empty(Config::$api_allowed_domains) or
+        in_array($_SERVER['HTTP_ORIGIN'], Config::$api_allowed_domains)
+    ) {
+        header('Access-Control-Allow-Origin: *');
```
AJDurant: potentially use the origin that was sent? seems to generally be recommended over * - even tho it is restricted already

Flump5281: I've seen both implementations used, personally. I don't particularly mind which we go for.

PR #405, opened 2015-03-12
Fix Mixclouder breakage due to API expecting POST request for setMeta

adfw:
As above.

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Swagger.php
+++ b/src/Classes/ServiceAPI/MyRadio_Swagger.php
@@ -303,6 +303,7 @@ class MyRadio_Swagger
      * 
      * Specify these with one or more @api decorators.
      * Defaults to GET only.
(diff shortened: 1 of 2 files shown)


AJDurant:
Surely this should be a change on mixclouder, set commands should really be POST requests

adfw:
Mixclouder would need much more refactoring compared to setting this attribute, as it has no concept of "What API request am I sending", it just sends them

CaptainHayashi:
Looks fine to me (well, about as fine as comments with non-obvious semantic value can be), but I'll defer to @AJDurant with regards to approval.

AJDurant:
It should be relatively easy to get mixclouder to use a post method though?
Any library that lets you send http will let you set the method.
I don't really like this as a 'fix'

adfw:
Fixed in mixclouder - uses post, tested working. This commit now addresses a documentation issue for the getOptionsAllow function.

PR #410, opened 2015-03-16
Add memcached provider, remove APC from setup checks

Flump5281:


src/Classes/MemcachedProvider.php
```diff
@@ -0,0 +1,179 @@
...
+ */
+
+namespace MyRadio;
+
+use \Memcached;
+
+/**
+ * MemcachedProvider provides in-memory caching for PHP resources to increase page load times
+ *
+ * MemcachedProvider was the first CacheProvider implementation in MyRadio. It enables Models to send cache commands to it
```
adfw: It wasn't the first :P

PR #417, opened 2015-04-12
Force PostgreSQL connection timezone

LloydW93:


LloydW93:
To summarise IRC discussion:
- [ ] Revert #416 
- [ ] Update MyRadio_Season::schedule to ensure incoming times are UTC (or potentially as part of reading values in MyRadio, I'm undecided there)
- [ ] Update schedule allocation pages to ensure they understand they are being given UTC and to convert appropriately when presenting to user

PR #418, opened 2015-04-12
Update editSeason.php

AJDurant:


LloydW93:
Why is it not the data $id?

AJDurant:
because the $data is only on form submitting. However this form for editing doesn't work anyway. and it's not linked from anywhere

PR #440, opened 2015-05-10
Fixes #423

### LINKED ISSUE #423: Jukebox returning to non-fancy 'default' weightings from midnight to 0100
Jukebox is having a nightly retreat to 'old ugly radio' mode, using the old weightings and playlists between 0000-0100. This appears to be the only time this happens. Have gone back over previous nights' scrobbles and it seems consistent.

LloydW93:


AJDurant:
banner_timeslot changes have not been reflected in the db

PR #451, opened 2015-05-26
Precise search for Similar Tracks

Eiron:
Discovered issues with some of the selections of Similar Tracks in appropriate playlists (e.g. *Nirvana - Lithium (Live)* when *Nirvana - Lithium* is available). Copying - hopefully - for similar functionality as other library searches.

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Track.php
+++ b/src/Classes/ServiceAPI/MyRadio_Track.php
@@ -774,17 +774,17 @@ class MyRadio_Track extends ServiceAPI {
                     $c = self::findByOptions(['title' => $r['name'],
                                 'artist' => $r['artist']['name'],
                                 'limit' => 1,
-                                'digitised' => true
+                                'digitised' => true,
                                 'precise' => true]);
                     //Try to find a not-so-exact match
-                    if (!empty($c)) {
+                    if (empty($c)) {
                         $c = self::findByOptions(['title' => $r['name'],
                                 'artist' => $r['artist']['name'],
                                 'limit' => 1,
-                                'digitised' => true
+                                'digitised' => true,
                                 'precise' => false]);
                     }
-                    //Still nothing
+                    //If match found, add track to Similar list
                     if (!empty($c)) {
                         $this->lastfm_similar[] = $c[0]->getID();
                     }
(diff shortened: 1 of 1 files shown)


URY-GitHub:
Can one of the admins verify this patch?

LloydW93:
Should be good once you resolve merge conflicts.

src/Classes/ServiceAPI/MyRadio_Track.php
```diff
@@ -770,10 +770,21 @@ public function getSimilar() {
             }
             foreach ($data['similartracks']['track'] as $r) {
                 if ($r['match'] >= 0.25) {
+                    //Try to find an exact match
                     $c = self::findByOptions(['title' => $r['name'],
                                 'artist' => $r['artist']['name'],
                                 'limit' => 1,
-                                'digitised' => true]);
+                                'digitised' => true
```
AJDurant: needs a comma to separate elements in array

src/Classes/ServiceAPI/MyRadio_Track.php
```diff
@@ -770,10 +770,21 @@ public function getSimilar() {
...
+                                'precise' => true]);
+                    //Try to find a not-so-exact match
+                    if (!empty($c)) {
```
AJDurant: if not empty is not the right condition to match. It probably will be empty if there are no matches, though that should be checked what findByOptions returns

src/Classes/ServiceAPI/MyRadio_Track.php
```diff
@@ -770,10 +770,21 @@ public function getSimilar() {
...
+                        $c = self::findByOptions(['title' => $r['name'],
+                                'artist' => $r['artist']['name'],
+                                'limit' => 1,
+                                'digitised' => true
+                                'precise' => false]);
+                    }
+                    //Still nothing
```
AJDurant: This comment is not true, the next condition only runs if there is something (it adds the found track to the array lastfm_similar)
Better comment would be something like: Add the track to the similar list if we have it

Eiron:
Yep. I see those errors. Not sure how they fill in. I copy-pasted the code, but must have made some weird errors somewhere along the line somehow. This version should fix those errors? Let me know if it needs any more repairs before it can be put live.

AJDurant:
you need to fix the merge conflicts, run 'git merge master' on your branch ad then fix anything that cannot be automatically resolved

AJDurant:
replaced by #452

PR #452, opened 2015-05-26
Precise search for getSimilar

Eiron:
I think this deals with the clashes. Hopefully good to go now.


URY-GitHub:
Can one of the admins verify this patch?

PR #454, opened 2015-05-29
fix request & allocate timezone issues

AJDurant:
Requesting a show was setting the requested time to the GMT time.
Allocation was compensating for this in the display, but not backend.
The timezone was then being re-compensated for on allocation, causing conflicts to happen with shows 2 hours previous.
Conflict end time params were wrong.


src/Classes/MyRadio/MyRadioFormField.php
```diff
@@ -784,7 +784,7 @@ private function convertTime($timeString)
          * YES, PHP IS RETARDED.
          */
         $timeString = str_replace('/', '-', $timeString);
-        $time = (int) strtotime($timeString) - date('Z');
+        $time = (int) strtotime($timeString);
```
LloydW93: No, the aim is for all server-side business logic to be UTC. I don't think this should be changed.

src/Classes/ServiceAPI/MyRadio_Season.php
```diff
@@ -716,8 +716,8 @@ public function getRequestedTimes()
 
     private function formatTimeHuman($time)
     {
-        $stime = gmdate(' H:i', $time['start_time'] + date('Z'));
-        $etime = gmdate('H:i', $time['start_time'] + $time['duration'] + date('Z'));
+        $stime = date(' H:i', $time['start_time']);
+        $etime = date('H:i', $time['start_time'] + $time['duration']);
```
LloydW93: I guess this is actually the same thing. I think I may have only done it like that to make it clear the localisation adjustment was intended.

src/Classes/ServiceAPI/MyRadio_Season.php
```diff
@@ -920,7 +920,7 @@ public function schedule($params)
                 $day_start = $start_day + (($i - 1) * 7 * 86400);
                 $show_time = $day_start + $req_time['start_time'];
 
-                $conflict = MyRadio_Scheduler::getScheduleConflict($day_start + $req_time['start_time'], $day_start + $start_time + $req_time['duration'] - 1);
+                $conflict = MyRadio_Scheduler::getScheduleConflict($show_time, $show_time + $req_time['duration']);
```
LloydW93: The -1 might need to stay in, have you tested it?

AJDurant:
okay, but the adjustments made elsewhere were working in local time, then being double adjusted. A show requested for 4pm (3pm in UTC) was being checked and clashing with a show at 2pm.
Perhaps I went the wrong way of 'fixing' it, but the scheduled time is correctly displayed everywhere now.

LloydW93:
Elsewhere?

src/Classes/ServiceAPI/MyRadio_Season.php
AJDurant: the -1 is done in MyRadio_Scheduler::getScheduleConflict so it seemed like it was being applied twice

LloydW93: (y)

AJDurant:
so the problem was a show was requested for 4pm. This requested time was stored as 15 hours in the db (3pm UTC) - however, then scheduling for 4pm it was being converted to 2pm and discovering a clash.
So if the field decoding and formatHumanTime are changed back, then more needs to be done on the bit that works out the actual start time.
Timeslots are still stored with time-zone though

LloydW93:
Hmm, sounds like I need to read around the code a little more to fully understand you, but sounds reasonable ish.

Timeslots are indeed still stored with time-zone, however at database connection startup the connection is negotiated to UTC now.

AJDurant:
The whole workflow of this is bad, I started planning out some changes, but haven't got much further at the moment

src/Classes/MyRadio/MyRadioFormField.php
AJDurant: What this line does is confuse things. It removes an hour from the duration field on the allocate form, and combined with other things, makes the displayed requested time different from the actual requested time.

LloydW93: Does this mean the displayed requested time (for the person scheduling) will always be UTC then?

src/Classes/ServiceAPI/MyRadio_Scheduler.php
```diff
@@ -152,21 +152,19 @@ public static function getTermDescr($termid)
...
-         * An extra hour is added here due to some issues with timezones and public.terms - some
-         * terms are set to start at 11pm Sunday instead of Midnight Monday. It's annoying because then we convert it back.
-         * If we didn't it's not the end of the world - the usage for this does not include time so just the date *should*
-         * be sufficient.
-         * @todo Fix terms database so it isn't silly.
-         */
-
-        return strtotime('Midnight '.date('d-m-Y', strtotime($result['start'])+3600));
+
+        return strtotime('Midnight ' . gmdate('d-m-Y', strtotime($result['start'])) . ' GMT');
```
LloydW93: :+1:

src/Classes/MyRadio/MyRadioFormField.php
AJDurant: yes and no, requested time is an offset from 1970-01-01 00:00:00 as it is only the hours and mins that are relevant - the date is calculated by whatever uses this, and it is added as an offset to the datetime

src/Classes/ServiceAPI/MyRadio_Scheduler.php
AJDurant: because specifying the timezone in two different ways is necessary in PHP...

AJDurant:
- [ ] test conflict detection
- [ ] test demos
- [ ] this banners
- [ ] this playlists

src/Classes/MyRadio/MyRadioFormField.php
LloydW93: Ah, yes of course.

Issue #453, opened 2015-05-28
API should use a "mixins" model to enable more granular control


LloydW93:
I'm currently imagining something along these lines:
- A permissions record, in addition to existing, has an api_mixin_auth table of the format (authid, class, mixin, effective_from, effective_to)
- toDataSource models are updated to have one more more "mixins" defined as strings. For example, users could have 'personal_data', 'training', 'officerships', and shows/seasons/timeslots could have 'buttons', 'credits'
- The existing 'full' option on toDataSource goes away as this is replaced by the mixins

Thoughts?

Issue #306, opened 2015-01-09
Remove get/set prefixes from API methods


LloydW93:
Probably requires an API v2

Instead, we should automatically use this information to determine whether a GET or POST is to be used, and enforce it.

Issue #305, opened 2015-01-09
Improve security of API credentials validation


LloydW93:
This will probably require a v2 API endpoint, but:

Currently, passwords are send for validation plaintext using the API. Although this is over HTTPS, if the API becomes more generally usable and a caching layer needs to go in, this could be a problem. We also have API logging currently disabled to prevent user/password tests being stored plaintext in the database, because that would be bad.

Some kind of client-side tokenisation needs to be conceived so that the password is sent using some form that the server is able to understand and test that doesn't expose the underlying password storage and ideally is not decodeable from log or error data.

Initially, I can think of something similar to the AWS signing method (http://docs.aws.amazon.com/general/latest/gr/sigv4_signing.html), although possibly with reduce complexity. We can add one or more fields that are things not logged or stored once the request has been processed.

For that to work, we'd also need to make the validation a POST request, which seems fair to me.

Example:
Username: abc123
Password: thisismypassword
APIKey: xxxxxxxx

1) Generate a random token (with a minimum length of say, 2048 characters)
2) Encrypt the payload (e.g. mcrypt_encrypt in PHP) with a salt of this string, potentially concatenated with other values such as a request timestamp, user agent, referrer...
3) POST the request to the server, with a sig=$the_random_string
4) Server rejects GET requests and the logging system does not store the sig parameter

Needs a lot more thinking through. Other options generally:
Log requests without payload data generally (could be annoying for debugging)
Have a blacklist of parameters not to log (problematic maintenance overhead)

PR #455, opened 2015-05-31
API v2

### LINKED ISSUE #121: API Key logging is disabled
The last log entries are from January.

I only noticed because I wanted to investigate a new spam user.

LloydW93:
Request for feedback and comments.

I aim to get #453 and #306 in this one. I'd also like to have a go at #305 to fix #121.

Note this currently has no authentication or authorisation, so using it is a bad idea.


LloydW93:
New table added:

CREATE TABLE myury.api_mixin_auth (
    api_mixin_auth_id SERIAL,
    class_name CHARACTER VARYING NOT NULL,
    mixin_name CHARACTER VARYING,
    typeid INT REFERENCES l_action(typeid)
);

Same general layout as any other auth granting table.

LloydW93:
Changes to the permissions model will require a complete cache wipe - otherwise you'll find Users can no longer Anything.

LloydW93:
Okay - pretty much there. However, the User endpoint now has no $full thing, so it's not entirely backwards compatible. Will hopefully be fine.

Might break the website though.

PR #457, opened 2015-06-02
Fixes #456. Attempt to add cache refreshing. Quite possibly not the correct thing to

### LINKED ISSUE #456: Assigning officership does not invoke cache update.

adfw:
do - may force an update before object exists?. Needs test.


LloydW93:
No.

LloydW93:
(as discussed on IRC, MyRadio_Profile does not exist. Profile needs to be imported in the namespace. getInstance, where implemented does not take Cache keys as a parameter, so I think the best approach would be a Profile::clearCache method)

PR #458, opened 2015-06-11
Fix photo uploads by removing fileinfo dependency

LordAro:


LloydW93:
Dependency checker in setup should be updated to reflect the change, if you're sure nothing else uses Fileinfo. If other stuff does use Fileinfo (which I get the feeling it does), then this PR make no difference, except potentially being slightly slower as it works out more information.

LordAro:
Good catch on the setup
Not used anywhere else, grepping for "finfo" or "fileinfo" brings up nothing of relevance
As for being slower... I hardly think it makes a difference - actually uploading the image is going to be the bottleneck here, and it's not exactly run often anyway

Issue #448, opened 2015-05-19
Explicit button on Track Uploader does nothing?


Eiron:
Least, those tracks I have checked as explicit are coming through as 'u'. One even upgraded itself to a 'y'. It was definitely not a 'y'.

AJDurant:
code looks correct, will test when time

LloydW93:
https://github.com/UniversityRadioYork/MyRadio/blob/master/src/Classes/Daemons/MyRadio_ExplicitDaemon.php#L43

May be part of the problem - things may be initially getting marked as unknown for some reason then this comes along and makes its guess, OR, findByOptions' clean filter is broken.

adfw:
https://github.com/UniversityRadioYork/MyRadio/blob/master/src/Controllers/NIPSWeb/confirm_central_upload.php#L17 could be the culprit - if it's not set, then it becomes NULL. Meaning https://github.com/UniversityRadioYork/MyRadio/blob/master/src/Classes/ServiceAPI/MyRadio_Track.php#L662 never becomes true. Although this is about Explicit going 'u' as well, could be something to check.

AJDurant:
@adfw that is the intended behavior, the explicit tick should only set explicit, not clean. So it will set explicit, or it will send null and be unknown. But the functionality is there in track creation so other methods can set tracks to be clean. See the discussion on #326

adfw:
Should be a code comment there then about the (current) dead code

AJDurant:
It's not dead code, and it's pretty obvious

adfw:
I've just shown that it's not clear that the fact it will *never* set clean to true, even though the code to set it true is present. Needs a note saying that the check for explicit === false is for future enhancements.

AJDurant:
That function *will* set clean to 'y' if it is asked to do so. That is not dead code.
The fact that the user-side functionality won't ask it to should not be documented in this function, as it has no relevance there. Perhaps on the user side.

adfw:
So documentation on the isset at https://github.com/UniversityRadioYork/MyRadio/blob/master/src/Controllers/NIPSWeb/confirm_central_upload.php#L17 then, since that prevents a 'y' getting set.

Eiron:
[...]

Issue #442, opened 2015-05-14
something weird with shows disappearing


AJDurant:
It seems, that if a person is credited to a show, and removed from that show, and then re-credited then the show sometimes appears and sometimes does not.

I guess, if this is the case, that it is due to the order in which SQL returns the credits, adding and removing people, however I'm not sure why this would affect other things - getCurrentAndNext returns null for title and description when this happens. The show does not show up in 'My Shows' or in 'Timeslot selection' but does in 'This Term/All Shows'

AJDurant:
fixed in 3c9014 in #482

Issue #437, opened 2015-05-10
fixing #429 lost the scrollbar for uploading tracks


AJDurant:
this should be added back, and the modal should probably be bigger anyway

Eiron:
Lost the scrollbars throughout Show Planner, generally.

AJDurant:
#429 was reverted

PR #459, opened 2015-06-11
Fix season owner check

LordAro:
Found looking at the logs. Not actually tested, done based on other usages of the function

Also, if there are any other *minor* fixes needed anywhere, let me know and i'll take a look

### DIFF
--- a/src/Controllers/Scheduler/editSeason.php
+++ b/src/Controllers/Scheduler/editSeason.php
@@ -26,7 +26,7 @@ if ($_SERVER['REQUEST_METHOD'] === 'POST') {
         $season = MyRadio_Season::getInstance($data['id']);
 
         //Check the user has permission to edit this show
-        if (!$season->isCurrentUserAnOwner()) {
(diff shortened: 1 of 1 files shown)


LloydW93:
https://github.com/UniversityRadioYork/MyRadio/issues/448
https://github.com/UniversityRadioYork/MyRadio/issues/446
https://github.com/UniversityRadioYork/MyRadio/issues/437

They should all be quick.

Also, what's the "problem" here?

LordAro:
"isCurrentUserAnOwner" is a member of MyRadio_Show, not MyRadio_Season

LloydW93:
Sorry, misremembered the DB schema. However, that's still technically not right. If a person is credited just for the duration of a specific episode, they're supposed to be an owner for that episode, but this checks if they are an owner for *right now*.

To be honest, the logic and planning around this was never finalised so there's some degree of undefined behaviour when you try to do complicated things with credits, which may be related to https://github.com/UniversityRadioYork/MyRadio/issues/442

Well volunteered :)

PR #462, opened 2015-06-16
Fix Show Planner scroll

LordAro:


LloydW93:
How'd you come up with 280px?

LordAro:
arbitrary :)

AJDurant:
Overflow hidden should be there, just not for the uploader

LordAro:
It broke more things than it fixed, i think it needs a different solution, if indeed it's needed at all, i believe decreasing the minimum height to such a level should fix the "jumpiness" of show planner described in the original issue. Certainly stuff going off the bottom of the screen (like progress bars, play buttons, etc) should never be unclickable just because your screen isn't big enough

LloydW93:
I guess this'll do for now, but tbh some proper responsive CSS should be done around height - below a certain screen height enable the scroll bar.

LloydW93:
If you're happy, I'm happy.

PR #480, opened 2015-07-13
Support datasource responses with mixins

LloydW93:
Some things, such as /api/v2/list/22/members, returns a set of dataSource'd objects from another class. Mixins are passed through correctly, but permissions were not properly verified.

This fixes that.

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_List.php
+++ b/src/Classes/ServiceAPI/MyRadio_List.php
@@ -314,7 +314,7 @@ class MyRadio_List extends ServiceAPI
     }
 
     /**
-     *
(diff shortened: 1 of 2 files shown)


LloydW93:
Todo: Come up with a way to make Swagger's documentation state that mixins for the response class are supported.

src/Classes/ServiceAPI/MyRadio_List.php
```diff
@@ -313,6 +313,10 @@ public static function getByName($str)
         }
     }
 
+    /**
+     *
```
LordAro: Could you not document this properly? :)

PR #482, opened 2015-07-30
Api functions

AJDurant:
updates to make the API more usable


src/Classes/ServiceAPI/MyRadio_Timeslot.php
```diff
@@ -348,6 +348,17 @@ public static function getMostListened($date = 0)
         return $top;
     }
 
+    public static function getFromSlug($slug)
```
LloydW93: I kinda feel like there should be a more general purpose search-by-meta function. If there isn't one already. And if there isn't one already, how did we get this far? :P

AJDurant: yeah probably, and this is pretty dirty as I didn't even limit it to the correct metadata type

schema/api.json
```diff
@@ -18,6 +18,8 @@
 	"\\MyRadio\\ServiceAPI\\MyRadio_Team": "team",
 	"\\MyRadio\\ServiceAPI\\MyRadio_TracklistItem": "tracklistItem",
 	"\\MyRadio\\ServiceAPI\\MyRadio_User": "user",
+    "\\MyRadio\\ServiceAPI\\MyRadio_Banner": "banner",
+    "\\MyRadio\\ServiceAPI\\MyRadio_BannerCampaign": "bannerCampaign",
```
LordAro: muh tabs

AJDurant: fixed

PR #496, opened 2015-09-03
lazy loading of payments fix

AJDurant:
I'm not actually sure why activateMemberThisYear exists... Maybe it would be useful as a permission block if it didn't take in the payment amount as a variable.


adfw:
I'd suggest also wrapping this in a check for early loop exit to be sure?

PR #498, opened 2015-09-03
getAll() with keys

AJDurant:
should fix the problems with getAllKeys() which is a fairly useless function:
"This is not an atomic operation, so it isn't a truly consistent snapshot of the keys at point in time. As memcache doesn't guarantee to return all keys you also cannot assume that all keys have been returned."
Every so often it would just return false when keys existed.

With this we keep a list of keys to be fetched then fetch them.


LloydW93:
:+1:

AJDurant:
- [ ] test on dev

PR #507, opened 2015-10-11
Improve Track/Tracklist performance

LloydW93:
Also fixes permissions on a cold/no cache.


src/Classes/Database.php
```diff
@@ -174,17 +173,19 @@ public function fetchAll($sql, $params = [])
...
      * @return Array The requested result row, or an empty array on failure
      * @throws MyRadioException
      */
     public function fetchOne($sql, $params = [])
     {
-        try {
-            $result = $this->query($sql, $params);
-        } catch (MyRadioException $e) {
-            return [];
+        if (!is_resource($sql)) {
```
AJDurant: if it is a resource then it will try to return $result which is undefined?

src/Classes/Database.php
```diff
@@ -174,20 +173,22 @@ public function fetchAll($sql, $params = [])
...
+            try {
+                $sql = $this->query($sql, $params);
+            } catch (MyRadioException $e) {
+                return [];
+            }
+        } else {
 
-        return pg_fetch_assoc($result);
+        return pg_fetch_assoc($sql);
```
AJDurant: this is, perhaps, more silly...

src/Classes/ServiceAPI/MyRadio_TracklistItem.php
```diff
@@ -227,9 +231,7 @@ private static function trackAmalgamator($result, $playlists = true)
 
             if ($playlists) {
                 $playlistobjs = iTones_Playlist::getPlaylistsWithTrack($trackobj);
-                foreach ($playlistobjs as $playlist) {
-                    $track['in_playlists'] .= $playlist->getTitle() . ', ';
-                }
+                $track['in_playlists'] = implode(', ', array_map(function($i){return $i->getTitle();}, $playlistobjs);
```
AJDurant: Parse error: syntax error, unexpected ';' in src/Classes/ServiceAPI/MyRadio_TracklistItem.php on line 234

LloydW93:
Much better.

AJDurant:
you changed `myury.stats.jukeboxplaycounter.js` without changing it to `myradio`

AJDurant:
you probably need to update the reference to that file, not just rename it...

LloydW93:
Apparently I did but didn't include them in the commit...

Issue #463, opened 2015-06-21
Review use of | raw filter in Templates


LloydW93:
https://gist.github.com/LloydW93/d36bb870b5013dd87e3c

Some of these (e.g. the Impersonator one) are definitely unneeded, whilst others (e.g. the Captcha one) definitely are. We should review this and remove those that are not needed.

I've already gone through some code (Form logic) and templates (base, scheduler, notices) today and removed multiple XSS attack vectors.

PR #516, opened 2015-10-26
this should fix up #463

AJDurant:
#463 still has raw in error and in news which might need doing too

### DIFF
--- a/src/Templates/parts/table-content.twig
+++ b/src/Templates/parts/table-content.twig
@@ -18,7 +18,7 @@
                 <a href="{{column.url}}" title="{{column.title}}"><span class="glyphicon glyphicon-{{column.value}}">&nbsp;</span></a>
         {% endif %}
       {% else %}
-                {{column | raw}}
+                {{column}}
      {% endif %}
               </td>
     {% endfor %}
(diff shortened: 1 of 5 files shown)


src/Templates/parts/table-content.twig
```diff
@@ -18,7 +18,7 @@
                 <a href="{{column.url}}" title="{{column.title}}"><span class="glyphicon glyphicon-{{column.value}}">&nbsp;</span></a>
         {% endif %}
       {% else %}
-                {{column | raw}}
+                {{column}}
```
LloydW93: I remember that this will break something, but I can't remember what. Helpful, I know.

AJDurant: I thought it would break icons or links, but it does not. The only thing I
could see is show descriptions, but I don't think that matters too much.
And even then raw is not a good filter, there should be a formatting only
safe filter, which is how the new front-end site works.

LloydW93: Yep, that would've been it, and that would be right. It comes back to the currently over-flexible TinyMCE

LloydW93:
Twig doesn't have a safe filter, so this'll do for now.

PR #517, opened 2015-10-26
reduce tinymce controls and formatting

AJDurant:


src/Public/js/myradio.form.js
```diff
@@ -356,8 +356,14 @@ window.MyRadioForm = {
...
             tinymce.init(
                 {
-                    selector: "textarea.myury-form-tinymce",
-                    plugins: "anchor autolink charmap code contextmenu fullscreen hr image link lists media tabfocus table wordcount",
+                    selector: "textarea.myradio-form-tinymce",
+                    plugins: "anchor autolink charmap code contextmenu hr image link lists media tabfocus table wordcount",
+                    media_alt_source: false,
+                    menubar: false,
+                    toolbar: "undo redo | styleselect | bold italic underline | alignleft aligncenter alignright alignjustify | bullist numlist outdent indent | link image media charmap | code",
+                    invalid_elements: "script",
```
LloydW93: Do we want to make <style> invalid too?

AJDurant: I was thinking about this, and some styling might be wanted, so I've blocked colour font-size background attributes, I wasn't sure whether to just remove style entirely or not.

LloydW93: It's probably more a business decision than a technical one, then. There's still no server side validation, and javascript: urls may well still work, but other than that seems vaguely sensible.

I'd also potentially look at the link policy - maybe they should always be target=_blank, or maybe with an exception for stuff on a certain domain?

PR #518, opened 2015-10-29
Add Sentry integration

wlcx:
This (MyRadioException) was the easiest thing to start with, not sure if there are any other useful things to hook into


AJDurant:
Could be worth adding user / session context into the logging (as session is already done in emailed reports)

LloydW93:
I think there's a Util function for generating that hopefully-without-passwords

PR #519, opened 2015-10-31
Js api

### LINKED ISSUE #385: Twig API macro should make requests go in AJAX with the right HTTP method

AJDurant:
fix #385 at least, in the only places it is used.

I'm not sure I'm completely happy with the twig macro, or the js function - but would like comments


LloydW93:
Merge plz :)

PR #520, opened 2015-11-04
Api mixin fun

wlcx:
This PR solves issues that were blocking aliasgen, namely not being able to pass the `personal_data` mixin for `MyRadio_User` 'through' api methods in classes that return nested data from `MyRadio_User`.
I have solved it by disabling checks for mixin names by default, and by somewhat hackily blindly passing the array of mixins through to these nested 'sub models'.

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Alias.php
+++ b/src/Classes/ServiceAPI/MyRadio_Alias.php
@@ -149,7 +149,7 @@ class MyRadio_Alias extends ServiceAPI
         $data = [
             'alias_id' => $this->getID(),
             'source' => $this->getSource(),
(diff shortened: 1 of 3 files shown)


LloydW93:
Ah, so I've just remembered that auth for mixins on the API is checked for the original called class...

AJDurant:
Also this would cause problems if there are mixins named the same thing

wlcx:
Yes it would. But then that is not solvable without changing the way mixins are named or requested.
They could be namespaced, e.g. `users.personal_data`, `officers.history` etc? Then they could be just passed to the models as per their namespace?

LloydW93:
As discussed, I think we shouldn't do this, and instead change the function used for fetching alias data.

PR #532, opened 2015-12-04
Fixes #321

### LINKED ISSUE #321: Manual Track Uploader can be overriden by duplicate detector
https://github.com/UniversityRadioYork/MyRadio/blob/master/src/Classes/ServiceAPI/MyRadio_Track.php#L598

Here, an album is automatically created based on the input.

https://github.com/UniversityRadioYork/MyRadio/blob/master/src/Classes/ServiceAPI/MyRadio_Track.php#L613

However, here any match for artist/album is used, and can override the input album.

LloydW93:
Fixes #321

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Track.php
+++ b/src/Classes/ServiceAPI/MyRadio_Track.php
@@ -642,14 +642,21 @@ class MyRadio_Track extends ServiceAPI
         // We need to rollback if something goes wrong later
         self::$db->query('BEGIN');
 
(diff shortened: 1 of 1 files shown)


CaptainHayashi:
Aha, that'd do it I guess

CaptainHayashi:
Have tested on myradio-dev with a track that was causing problems for me earlier (of the mid-show sweary I REALLY WANT TO PLAY THIS kind).  Seems to work.  Going to bite on the merge.

Issue #536, opened 2016-01-04
show planner library selection text boxes


adfw:
Improvements to the library toolbar with PR #531 broke the search boxes.
Two-stage PR: 1 to rollback the fixes quickly, and another to actually fix the first thing without breaking... See issue #274

mstratford:
Something possibly to do with #649...

mstratford:
Fixed with #649 & Fixes #274

Issue #274, opened 2014-12-28
Investigate inset on artist search in show planner


LloydW93:

mstratford:
Fixed by #649

PR #537, opened 2016-01-04
Hotfix to #536 - will probably result in #274 reappearing

adfw:
Replace the new show/hide JS with the existing. Better than having no
search boxes at all.

**Do not delete branch after merge** -- JS fix to #274 needs reapplying without breakage this time


CaptainHayashi:
Seems ok after a bit of honestly not that exhaustive testing (flicking between various libraries and dragging a few songs in and out of then)

PR #552, opened 2016-03-01
Update yusu API endpoint

adfw:
This should probably be a configurable item, since it seems to change all the time...

### DIFF
--- a/src/Classes/Config.php
+++ b/src/Classes/Config.php
@@ -468,6 +468,12 @@ final class Config
     public static $yusu_api_key;
 
     /**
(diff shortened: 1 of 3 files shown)


AJDurant:
These should be in Config

AJDurant:
you need to actually create the variable in the config class

AJDurant:
The only other thing is how do we handle when the Config is not set. Cause things in local config might not be and that shouldn't cause stuff to break too much. However I'm not sure it's worth it - it shouldn't ever not me set, and it would fail anyway if it doesn't return data.

PR #561, opened 2016-05-24
Fix session errors due to not correctly creating null session

LloydW93:


src/Classes/MyRadio/MyRadioSession.php
```diff
@@ -19,7 +19,8 @@ public static function factory()
             session_write_close();
         }
 
-        return new self();
+        $class = get_called_class();
```
GrahamCampbell: It would be better to just use late static binding and have `return new static;`, rather than this hacky call to work out the parent class.

PR #563, opened 2016-05-26
unrestrict future shows

AJDurant:
Currently you can't view/work on shows that are in the future, this is a problem if you need to plan a future show block


LloydW93:
I believe the original intention with this was to prevent users who were just added from working on shows for before they were added. Why *is* effective_from in the future here, if it's a new show or you're the owner you should have effective permissions straight away.

AJDurant:
Because people are only presenting part of 101 so have short effective
times so the credits on the website show up nicely. Ideally the effective
from/to times would be applied to timeslots so that you can always access
stuff your credited for, but not the rest but that was going to be a fairly
major refactor that I didn't want to do last night, this is relatively
simple and shouldn't give overt access.

LloydW93:
I didn't think it was a problem, just wanted to say the thought.

PR #565, opened 2016-05-26
WIP: Make API use timestamps everywhere

### LINKED ISSUE #357: Datatables Date sort dodgy
From phab

LordAro:
Makes the API return unix timestamps everywhere, instead of varying amounts of precision and formats

Will break things.

To check:
- [ ] banner campaigns
- [ ] quote edit form
- [ ] albums
- [ ] nipsweb managed items, whatever they are
- [ ] Everything else that uses the API
- [ ] news items

Fixes #357. Probably.

### DIFF
--- a/src/Classes/MyRadio/CoreUtils.php
+++ b/src/Classes/MyRadio/CoreUtils.php
@@ -496,7 +496,12 @@ class CoreUtils
         $return = [];
         $return[] = ['Timestamp', 'Errors per request', 'Exceptions per request', 'Queries per request'];
         foreach ($result as $row) {
-            $return[] = [self::happyTime($row['timestamp'], true, false), (int) $row['errors'], (int) $row['exceptions'], (int) $row['queries']];
+            $return[] = [
+                self::happyTime($row['timestamp'], true, false),
+                (int) $row['errors'],
+                (int) $row['exceptions'],
+                (int) $row['queries']
+            ];
         }
 
         return $return;
--- a/src/Classes/MyRadio/CoreUtils.php
+++ b/src/Classes/MyRadio/CoreUtils.php
@@ -148,19 +148,26 @@ class CoreUtils
 
     /**
      * Returns a postgresql-formatted timestamp.
-     *
-     * @param int $time The time to get the timestamp for. Default right now.
-     *
+     * @param int $time The time to get the timestamp for. Defaults to "now".
      * @return string a timestamp
      * @assert (30) == '1970-01-01 00:00:30'
      */
     public static function getTimestamp($time = null)
     {
-        if ($time === null) {
-            $time = time();
-        }
-
+        if (!isset($time)) $time = time();
         return gmdate('Y-m-d H:i:s+00', $time);
+    }
+
+    /**
+     * Returns a postgresql-formatted duration.
+     * @param int $time The time to get the duration for. Defaults to "now".
+     * @return string a duration
+     * @assert (4242) == '01:10:42'
+     */
+    public static function getDuration($time = null)
+    {
+        if (!isset($time)) $time = time();
+        return gmdate('H:i:s', $time);
     }
 
     /**
--- a/src/Classes/MyRadio/CoreUtils.php
+++ b/src/Classes/MyRadio/CoreUtils.php
@@ -121,7 +121,8 @@ class CoreUtils
      */
     public static function happyTime($timestring, $time = true, $date = true)
     {
-        return date(($date ? 'd/m/Y' : '').($time && $date ? ' ' : '').($time ? 'H:i' : ''), is_numeric($timestring) ? $timestring : strtotime($timestring));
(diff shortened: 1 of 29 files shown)


src/Classes/MyRadio/CoreUtils.php
```diff
@@ -148,22 +149,29 @@ public static function intToTime($int)
...
      * @return string a timestamp
      * @assert (30) == '1970-01-01 00:00:30'
      */
     public static function getTimestamp($time = null)
     {
-        if ($time === null) {
-            $time = time();
-        }
-
+        if (!isset($time)) $time = time();
```
LloydW93: gmdate should be doing this itself?

src/Classes/NIPSWeb/NIPSWeb_ManagedItem.php
```diff
@@ -71,9 +72,9 @@ protected function __construct($resid, $playlistref = null)
                 );
         $this->folder = $result['folder'];
         $this->title = $result['title'];
-        $this->length = strtotime('1970-01-01 '.$result['length']);
+        $this->length = $result['length'] ?: 0;
```
LloydW93: Show Planner my grumble at these

src/Classes/MyRadio/CoreUtils.php
LordAro: `gmdate(null);` results in an empty string

src/Classes/NIPSWeb/NIPSWeb_ManagedItem.php
LordAro: should be no different at all - strtotime just returns the unix timestamp, and I've modified the query to return epoch

LordAro:
Set up on /myradio-lordaro . There seem to be some weird 404s with no error messages anywhere...

src/Classes/MyRadio/CoreUtils.php
AJDurant: no inline if statements

src/Classes/MyRadio/MyRadio_Availability.php
```diff
@@ -113,21 +116,24 @@ protected function __construct($id, $result = null)
 
         $this->created_by = MyRadio_User::getInstance($result['memberid']);
         $this->approved_by = empty($result['approvedid']) ? null : MyRadio_User::getInstance($result['approvedid']);
-        $this->effective_from = strtotime($result['effective_from']);
-        $this->effective_to = empty($result['effective_to']) ? null : strtotime($result['effective_to']);
+        $this->effective_from = $result['effective_from'];
+        $this->effective_to = $result['effective_to'] ?: 0;
```
AJDurant: this may break things, that that is effectively saying effective_to is 1970

src/Classes/ServiceAPI/MyRadio_Album.php
```diff
@@ -81,13 +86,13 @@ protected function __construct($recordid)
         $this->media = $result['media_descr'];
         $this->format = $result['format_descr'];
         $this->record_label = $result['recordlabel'];
-        $this->date_added = strtotime($result['dateadded']);
-        $this->date_released = strtotime($result['datereleased']);
+        $this->date_added = (int) $result['dateadded'];
```
AJDurant: I think our db wrapper should do type conversions automatically

src/Classes/MyRadio/MyRadio_Availability.php
LordAro: And what would 'null' do? The same, afaict

src/Classes/ServiceAPI/MyRadio_Album.php
LordAro: ahahaha. OK, sure

LordAro: More helpfully, does the pg lib even return something approximating types?

LloydW93: http://php.net/manual/en/function.pg-field-type.php

Oh.

I have genuinely never thought to ask that question.

LordAro: :)
I'd suggest out of scope for this prq anyway

wlcx:
As a suggestion, ISO8601 tends to be prefered for APIs over unix timestamps

LordAro:
Yes, this should be rewritten to use ISO8601

MattWindsor91:
From a myradio-go perspective, it'd be nice to have this done, and a more robust way of handling things like "The End of Time" in place.

LordAro:
iirc myradio-go was the original reason this came about. I've since come to
the conclusion that ISO8601 timestamps would be more helpful than raw
timestamps

Would indeed be good to get this done though, even if it's basically a
rewrite (of the prq) at this point

LordAro:
Yeah ok, this is never going to get fixed in its current form. Closing.

PR #566, opened 2016-05-29
check timeslot permissions for the timeslot, making owners generic

AJDurant:
This also prevents access to non credited timeslots within a show, though they will still be visible in the list.

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Creditable.php
+++ b/src/Classes/ServiceAPI/MyRadio_Creditable.php
@@ -16,6 +16,7 @@ use MyRadio\ServiceAPI\MyRadio_User;
  */
 trait MyRadio_Creditable
 {
+    protected $owner;
     protected $credits = [];
     protected static $credit_names;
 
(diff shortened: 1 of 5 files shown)


src/Classes/ServiceAPI/MyRadio_Creditable.php
```diff
@@ -75,6 +76,24 @@ public function getCreditObjects($parent = null)
     }
 
     /**
+     * Checks the current user is in the credits for the creditable item
+     * @return boolean The user is an owner
+     */
+    public function isCurrentUserAnOwner()
```
adfw: Typo, should be 'isCurrentUserAndOwner' ?

AJDurant: no, it is asking 'is the current user an owner' this name has not changed

adfw: Ok, the function comment is slightly ambiguous but not a problem.

src/Classes/ServiceAPI/MyRadio_Show.php
```diff
@@ -586,20 +586,6 @@ public function getGenre()
         return isset($this->genres[0]) ? $this->genres[0] : null;
     }
 
-    public function isCurrentUserAnOwner()
```
adfw: Have you checked for all uses of this before moving it to a different class?

AJDurant: it's not in a different class, it's in the interface

adfw: This appears to be in MyRadio_Show, and it's moved to MyRadio_Creditable? Or is this not a problem?

AJDurant: MyRadio_{Show,Season,Timeslot} are instances of MyRadio_Metadata_Common which use MyRadio_Creditable

PR #569, opened 2016-09-06
Add ability to leave demos

### LINKED ISSUE #503: Cannot back out of studio training (nee demo)
Issue a new member had this week - they found they couldn't make the slot they signed up for, but had no way to deselect / back out of it. Required manual database change (deletion of row in schedule.show_credit)

wlcx:
Add ability to leave demos.
This involved reshuffling some logic from MyRadio_Demo to the controller, so that the link could change from join to leave as appropriate. Bonus - no join link is shown if the demo is full.
Resolves #503.


LordAro:
👍

PR #570, opened 2016-09-20
Update YUSU membership payment link

wlcx:


LordAro:
Don't suppose you could config it? :>

wlcx:
no u

wlcx:
no u

wlcx:
That's way more convoluted though? Why do config->injected into javascript->some javascript somewhere which loads it into the DOM when we have a perfectly good templating system? This is more explicit and doesn't require JS which is always a plus in my view.

LordAro:
because it's how it's done? by all means fix it in future, but it should be consistent with what we've got at the moment

wlcx:
Ok, what ware you talking about here? Which js config file?

LordAro:
My bad, i don't mean javascript at all, i mean the twig config object - https://github.com/UniversityRadioYork/MyRadio/blob/master/src/Classes/MyRadioTwig.php#L64 & https://github.com/UniversityRadioYork/MyRadio/blob/master/src/Classes/Config.php#L708

wlcx:
Ok that makes more sense :p

wlcx:
Updated

PR #571, opened 2016-09-20
Return `receive_email` by default for Member

ChrisTheBaron:
This is what was breaking aliasgen.

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_User.php
+++ b/src/Classes/ServiceAPI/MyRadio_User.php
@@ -2140,7 +2140,6 @@ class MyRadio_User extends ServiceAPI implements APICaller
                 $data['paid'] = $this->getAllPayments();
                 $data['locked'] = $this->getAccountLocked();
                 $data['college'] = $this->getCollege();
-                $data['receive_email'] = $this->getReceiveEmail();
                 $data['email'] = $this->getEmail();
                 $data['phone'] = $this->getPhone();
                 $data['eduroam'] = $this->getEduroam();
@@ -2161,6 +2160,7 @@ class MyRadio_User extends ServiceAPI implements APICaller
             'sex' => $this->getSex(),
             'public_email' => $this->getPublicEmail(),
             'url' => $this->getURL(),
+            'receive_email' => $this->getReceiveEmail(),
         ];
 
         $data['photo'] = $this->getProfilePhoto() === null ?


wlcx:
Squash commits plz :p

wlcx:
Also receive_email is part of the personal data mixin (see MyRadio_User.php Line 2143), so are you just not calling that mixin in your api calls?

ChrisTheBaron:
@wlcx calling `/officer/allofficerpositions` or `/team/%d/headpositions` with the mixin `personal_data` gets:
```
{
  "status": "FAIL",
  "payload": "Unsupported mixin personal_data",
  "time": "0.039237"
}
```
and calling `/alias/allaliases` with it doesn't change the data returned.

wlcx:
Ok that's a separate issue in that case - we should work out why that mixin doesn't work

AJDurant:
The mixin is not part of aliases, it is part of an internal call

LordAro:
Looks like the issue here is that officers don't have a concept of 'personal_data' mixin, only users have that:-
https://github.com/UniversityRadioYork/MyRadio/blob/master/src/Classes/ServiceAPI/MyRadio_User.php#L2139
https://github.com/UniversityRadioYork/MyRadio/blob/master/src/Classes/ServiceAPI/MyRadio_Officer.php#L774

ChrisTheBaron:
So suggestions on what to do? Make the mixin work for Officers? I don't know how mixins are meant to work either @LordAro :P

AJDurant:
What is the actual API call that is having problems?

ChrisTheBaron:
> calling /officer/allofficerpositions or /team/%d/headpositions with the mixin personal_data gets:
>
> {
>  "status": "FAIL",
>  "payload": "Unsupported mixin personal_data",
>  "time": "0.039237"
> }
> and calling /alias/allaliases with it doesn't change the data returned.

All these endpoints are called within aliasgen and the new alias-go.

AJDurant:
Right, so you can either use the data you get from those calls to make more
detailed calls to /user with the mixin. Or implement a mixin for those
calls that then uses the personal_data one when it gets the users
internally.
Or implement some kind of mixin pass-through which I remember discussing
with @LloydW at some point.

ChrisTheBaron:
@AJDurant I will probably do the first one for now, that will be easier :stuck_out_tongue_winking_eye:

AJDurant:
That will be a massive amount of API calls

ChrisTheBaron:
Can't I just get all members in one call?

AJDurant:
Maybe, but you don't want all members, you want the members info that
relates to the members returned by the officer calls

ChrisTheBaron:
If I get all members at once then I can use the information at any point. It's not only the Officers I need the `recieve_email` field for.

ChrisTheBaron:
Can somebody do a review of this please

LordAro:
Assigned to @AJDurant because he has a better idea of the direction the mixins are supposed to go

(ignore my approval :) )

wlcx:
Please squish commits :p

ChrisTheBaron:
@wlcx I thought github gave you the option to squash pr's when you merge them?

wlcx:
true dat, though it then doesn't use a merge commit, which can still be preferable. IMO it's always worth neatening up a branch, even if just to make reviewing easier.

ChrisTheBaron:
@wlcx I have squished, imo commits should be small units of work, but this is not the time or place for that discussion :smile:

wlcx:
Absolutely, would you not say this is a unit of work then? I would!

ChrisTheBaron:
I would say this is the culmination of a few units of work :stuck_out_tongue:

LordAro:
It should be individual units of work, but those units are not time
related. So if you revert something you did earlier, you should get rid of
commits referencing the change

But, this is not a discussion for here, go away :p

AJDurant:
Okay, you're kind of mixing some code designed to be generic into something fairly specific.

When you're getting Aliases are you likely to need any of the other mixins, or ever want them without the receive_email attribute? If not, then you don't need code in there to enable it, just always ask for the things you always need.

When calling officers you may want to be able to get different data, idk.

The other thing is whether receive_email is actually a personal_data item or not, it doesn't seem like a particularly private thing to me, and opening up the rest of that grouping just to get access to the one attribute doesn't seem great.

ChrisTheBaron:
> The other thing is whether receive_email is actually a personal_data item or not, it doesn't seem like a particularly private thing to me, and opening up the rest of that grouping just to get access to the one attribute doesn't seem great.

Just so I'm clear on this @AJDurant, you're suggesting to return `receive_email` by default for a member?

AJDurant:
yes, possibly. I think that discussion should be had

PR #576, opened 2016-10-08
Don't try to tracklist demos

LordAro:


adfw:
I've just realised an issue with this -- in the case that the studio is accidentally put on air, we *do* want demo slots to be tracklisted. I think this should be handled alternatively. The formatting change is good though.

adfw:
Actually, never mind, this will still tracklist, but won't set the timeslotid. So probably fine. I would recommend actually testing this theory in a demo slot though.

schema/base.sql
```diff
@@ -24,97 +24,119 @@ CREATE SCHEMA webcam;
 CREATE SCHEMA website;
 COMMENT ON SCHEMA website IS 'Collection of data relating to the operation of the public-facing website.';
 CREATE EXTENSION IF NOT EXISTS plpgsql WITH SCHEMA pg_catalog;
-CREATE FUNCTION bapstotracklist() RETURNS trigger
-    LANGUAGE plpgsql
-    AS $$DECLARE
-audid INTEGER;
+CREATE FUNCTION bapstotracklist() RETURNS TRIGGER AS $$
```
adfw: lacks $$LANGUAGE plpgsql prelude

LordAro: This was intentional - I moved it to the bottom of the functions, as per the examples on https://www.postgresql.org/docs/9.5/static/plpgsql-structure.html

adfw: Ok, missed that one, misread the sequence at the bottom of the diff as another function.

AJDurant:
These sql files are used for database setup, and don;t change anything directly, has the trigger been changed in our db? If so, I'm happy to merge as @adfw has reviewed.

LordAro:
Nothing's been changed yet

AJDurant:
Is the dB trigger going to be updated?

PR #577, opened 2016-10-09
Make MyRadio scale quite a bit better when viewed on a mobile device

LordAro:
Changes this:

to this:

I think this is much nicer :)


AJDurant:
It was always supposed to do this...

Issue #583, opened 2016-10-20
Add link to URY website in MyRadio navbar


ChrisTheBaron:
Something like this would be nice:

![pasted image at 2016_10_20 19_49](https://cloud.githubusercontent.com/assets/6439999/19573507/7027deb6-96fe-11e6-84c2-2ce0f5a71724.png)

TomBellis:
Would it go in: MyRadio/src/Templates/parts/nav.twig

LordAro:
I'd think so, but it'd want to be configurable (from php's config file) as well
Which'll make it a bit more fun

TomBellis:
@LordAro not too familiar with php. What do you mean by configurable?

TomBellis:
So can we add:  
`<a class="navbar-brand" href="//ury.org.uk/">ury.org.uk</a>`
to the 
`MyRadio/src/Templates/parts/nav.twig`

AJDurant:
The url is in the config and should be read from there. I can't remember how you do that from Twig

Brookke:
I am confused, why would it be hard coded that the link title is ury.org.uk when the href should be set in config? If the href was to change to ury.radio wouldn't the link title want to change too?

AJDurant:
the url is already in the config file

PR #584, opened 2016-10-21
Add home page link to nav bar

Brookke:
Not sure if the var name homeurl is descriptive enough?

Closes: #583

### DIFF
--- a/src/Classes/Config.php
+++ b/src/Classes/Config.php
@@ -80,6 +80,13 @@ final class Config
     public static $schedule_url = '//ury.org.uk/schedule';
 
     /**
+     * The base URL of the radio home pages.
+     *
+     * @var string
+     */
+    public static $home_url = '//ury.org.uk/';
+
+    /**
      * Whether nice URL rewrites are enabled
      * If true, then urls will be myury/[module]/[action]
      * If false, then urls will be myury/?module=[module]&action=[action].
--- a/src/Classes/Config.php
+++ b/src/Classes/Config.php
@@ -84,7 +84,7 @@ final class Config
      *
      * @var string
      */
-    public static $home_url = '//ury.org.uk/';
+    public static $website_url = '//ury.org.uk/';
 
     /**
      * Whether nice URL rewrites are enabled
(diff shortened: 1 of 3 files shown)


src/Templates/parts/nav.twig
```diff
@@ -10,6 +10,7 @@
       </button>
       {% endif %}
       <a class="navbar-brand" href="{{baseurl}}">MyRadio</a>
+      <a class="navbar-brand" href="{{homeurl}}">ury.org.uk</a>
```
LordAro: Not really much point in making it configurable if you just hardcode the text anyway?

src/Classes/Config.php
```diff
@@ -80,6 +80,13 @@
     public static $schedule_url = '//ury.org.uk/schedule';
 
     /**
+     * The base URL of the radio home pages.
+     *
+     * @var string
+     */
+    public static $home_url = '//ury.org.uk/';
```
AJDurant: Config variables like this should really be declared here and set in the local config.

src/Templates/parts/nav.twig
AJDurant: It shouldn't have 'navbar-brand' read the Bootstrap docs.
And it should probably go into the hamburger menu on small screens.

Brookke: @LordAro I did mention that, I can also create a var for the name.

@AJDurant I will sort that now

AJDurant: @Brookke you've already got `$short_name` and `$long_name` in the config you could use (probably shortname). Maybe '`short_name` Website' or something.

src/Classes/Config.php
Brookke: @AJDurant How do I go about setting that in the local config? Ie how is that generated?

Brookke:
I think I have done everything asked except the local config thing, however it doesn't seem like the short name or long name are set in the local config either?

AJDurant:
Yeah, a load of stuff should be moved there. It's not a major issue. I just checked it out on -dev looks good 👍

Brookke:
Thanks @AJDurant 😄

PR #586, opened 2016-10-22
update tinymce, stop the extra html content, tidy text emails

AJDurant:
This will require running `composer update` after pulling.
Also the daemons will need restarting


src/Templates/form.twig
```diff
@@ -37,7 +37,7 @@
 {{ parent() }}
 {% include 'parts/table-footer.twig' %}
 <script type="text/javascript" src="{{baseurl}}js/vendor/jquery.validate.min.js"></script>
-<script type="text/javascript" src="{{baseurl}}js/vendor/tinymce.min.js"></script>
+<script src="//cdn.tinymce.com/4/tinymce.min.js"></script>
```
LordAro: In light of recent internet issues, do you think it's wise to use a CDN for this? Especially for just one, rather than the rest of them as well

AJDurant: It seemed like a big bunch of stuff to pointlessly store in our repo. Idk, we used to use CDNs for everything, I can't remember why we moved off them

AJDurant:
We can script something to remove them, shouldn't be too hard, but it's not worth doing it before this cause as soon as someone edits it it will go back in.

src/Templates/form.twig
AJDurant: Also, if this doesn't load then you just have a textbox, it doesn't break usage

adfw: A local copy would be preferable to me also, mostly because MyRadio currently does a very good job at having no external dependencies, which would be nice to keep for all the reasons above.

AJDurant:
Yeah they are fair comments. I was going to, and was just testing the new
version using the CDN, then I thought why bother to download it. I guess we
could possibly install it with composer, not sure how that works though -
or move our fontend dependencies to bower.

LordAro:
Thoughts on making tinymce do markdown?

AJDurant:
That would require more post processing. This update should fix pretty much
all the issues we have with it at the moment. It just needs to include the
local copy rather than use the cdn

mstratford:
Can't we just put the file in the Public JS folder and be done? Unless there is something I am missing that requires composer/bower? If so, I can easily do that! :)

LloydW93:
Yep, sticking it in public/js/vendor is what we're talking about here.

mstratford:
Right, the latest tinyMCE version is now locally included :)

mstratford:
Erm, I've now updated the file to the local one, it is now adding a <div> parent every time?

LordAro:
@AJDurant just in case you have any input, have an assignment

mstratford:
After testing this, TinyMCE is now behaving, but is has been suggested that https://github.com/UniversityRadioYork/MyRadio/blob/master/src/Classes/MyRadio/MyRadioFormField.php#L584 is adding root level <div> on every submission.

AJDurant:
Yeah as far as I can recall it has to have some element wrapping stuff to
make it work correctly and not explode all the content everywhere. It being
a div is safe cause it won't cause the problem of rewriting to a different
tag, or some such. But I would like to check the branch before merging to
make sure. Also I think I changed composer stuff in my commit which makes
the deployment slightly different.

AJDurant:
So the <div> wrapper is needed because DOMDocument needs a root node, otherwise it mutilates it's input. I thought a div would be transparent enough to not matter, but it does, so it now gets stripped off again.

AJDurant:
The db needs cleaning: every instance of `<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN" "http://www.w3.org/TR/REC-html40/loose.dtd">
<html><body><p>` and `</body></html>` and `<p>&nbsp;</p>` needs removing from the string.

src/Classes/MyRadio/MyRadioFormField.php
```diff
@@ -613,14 +613,16 @@ public function readValue($prefix)
...
+                $dom->loadHtml("<div>$_REQUEST[$name]</div>", LIBXML_HTML_NOIMPLIED | LIBXML_HTML_NODEFDTD);
 
                 $xpath = new \DOMXPath($dom);
                 while ($node = $xpath->query('//script')->item(0)) {
                     $node->parentNode->removeChild($node);
                 }
 
-                return $dom->saveHTML();
+                // Strip the <div> ... </div> nodes back off
+                return substr(trim($dom->saveHTML()), 5, -6);
```
LordAro: This is scary. Is there no better way of removing the root/getting the child node?

src/Public/js/myradio.form.js
```diff
@@ -356,11 +356,16 @@ var MyRadioForm = {
         media_alt_source: false,
         menubar: false,
         toolbar: "undo redo | styleselect | bold italic underline | removeformat | alignleft aligncenter alignright alignjustify | bullist numlist outdent indent | link image media charmap | visualblocks code preview",
+                    element_format : 'html',
```
LordAro: formatting

LordAro:
Several of the tinymce files seem to have been made executable (775). Not sure where this came from, but i think it'd be better if they weren't... there's no need, regardless

src/Classes/MyRadio/MyRadioFormField.php
AJDurant: Nope, it's not designed to be partial HTML/XML. The line that adds the wrapper garuntees this though

src/Public/js/myradio.form.js
AJDurant: Weird, must have been something with the merge conflict

AJDurant:
I can't fix the permissions thing on Windows. It all went a bit weird when I tried yesterday, I think it happened when it was added though.

AJDurant:
@LordAro Fixed permissions and the formatting in the js

Issue #585, opened 2016-10-21
Update Nav bar


Brookke:
Features:
- logo (myRadio specific)
- Brand Colour

something like below
![screen shot 2016-10-22 at 00 38 13](https://cloud.githubusercontent.com/assets/10200358/19615075/c8dd1cb6-97f2-11e6-83c0-dcf209d74367.png)

PR #587, opened 2016-10-22
Navbar styling update

Brookke:
- Added logo
- set colour

See what you think

Closes: #585


Brookke:
@ChrisTheBaron @sophiekinley can you take a look at this please and let me know what you think please 😄

LordAro:
Did you check that the burger menu still works on mobiles?

Brookke:
screen shot 2016-10-24 at 19 44 54

Brookke:
I can't create an account on my local my radio so I can't fully test. If someone could explain or test this for me that'd be great 👍  Thanks

LordAro:
It looks fine, but I'm inclined to agree with @LloydW93 on this - MyRadio isn't the website, so it shouldn't look the same. I'm all for a reskin if you feel that's necessary though

LordAro:
Gonna close this until we can come to some conclusion over styling. Might even want to talk to DCM or others :o

mstratford:
Just to comment here, tried it on my local account, however you managed to embed the text into the SVG image, the font doesn't carry over. Defaults to times new roman (or similar).

AJDurant:
If you use a font in an svg it had to be available to the end user, or you
need to convert it to a path. The font used to be in the stylesheet, it may
just not match the right name.

Issue #354, opened 2015-02-11
Track Edit Tool


LordAro:
An option to edit tracks in the Central Database including lyric cleanliness would be useful.

"There's a basic edit form now, but could do with more options." There is?

From T19

PR #591, opened 2016-10-28
Edit track form and functionality

AJDurant:
This doesn't currently have an easy way to access it, but I'm not sure when I'll be able to finish that and it would probably still be helpful to have now. you can access with editTrack?trackid=XXXX

### DIFF
--- a/src/Classes/MyRadio/MyRadioFormField.php
+++ b/src/Classes/MyRadio/MyRadioFormField.php
@@ -527,6 +527,19 @@ class MyRadioFormField
                 $options['trackname'] = $this->value->getTitle();
                 $value = $this->value->getID();
             }
+        } elseif (($this->type === self::TYPE_ALBUM) && !empty($this->value)) {
+            if (is_array($this->value)) { //Deal with TABULARSETs
+                foreach ($this->value as $k => $v) {
+                    if (empty($v)) {
+                        continue;
+                    }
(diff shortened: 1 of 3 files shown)


mstratford:
Fixes #354

mstratford:
If the track is loaded into Show Planner, then after After changing a track using this form, the outputs of the Central Music Library do show up when searching for the new track name, but the old track name is the one listed.
http://imgur.com/a/Wz59T

mstratford:
Also, before this goes available to all (with permissions), you may want to make it play nice with an empty trackid :)

AJDurant:
Hehe that first one is fun, there'll be a dirty cache entry.

LloydW93:
Could be related to #612

mstratford:
Just tried on MyRadio-dev, doesn't look like the Caching problem can be replicated, so looks ready to merge.

mstratford:
Seems ready to merge.

AJDurant:
The cache may be (and I'm pretty sure is) turned off on the -dev branch

mstratford:
I'll put it on my todo list 👍 I'll get round to it sooner rather than later.

AJDurant:
I think someone asked me if I approve this, I do but I can't formally review on here because it's my PR

mstratford:
There you go 👍

PR #595, opened 2016-10-31
Only show lists that you can actually subscribe or unsubscribe from -…

### LINKED ISSUE #593: Only show mailing lists that members are subscribed to, or can subscribe to
Basically hide any mailing lists on https://ury.org.uk/myradio/Mail/ that members shouldn't be able to subscribe to.

> But yeah something along the lines of: is list subscribe-able OR is member of list (you should still show up on that even if you are unsubscribed) - the generated lists are stored queries.

Will eliminate issues like: https://github.com/UniversityRadioYork/Issues/issues/55

AJDurant:
… fix #593 

This could probably be done slightly better, but give you an idea of what kind of thing to do.

Also some lists should probably be restricted somehow. With the current system it would be making them autolists, then dumping everyone into the opt-out automatically.

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_List.php
+++ b/src/Classes/ServiceAPI/MyRadio_List.php
@@ -67,6 +67,13 @@ class MyRadio_List extends ServiceAPI
     private $members = [];
 
     /**
+     * This is the set of members have opted out of this 'optin' list.
+     *
+     * @var int[]
+     */
+    private $nonMembers = [];
+
+    /**
      * Initialised on first request, stores an archive of all the email IDs
      * sent to this list.
      *
@@ -106,12 +113,22 @@ class MyRadio_List extends ServiceAPI
                 (SELECT memberid FROM mail_subscription WHERE listid=$1)',
                 [$listid]
             );
+            $this->nonMembers = self::$db->fetchColumn(
+                'SELECT memberid FROM mail_subscription WHERE listid=$1',
+                [$listid]
+            );
         }
         $this->members = array_map(
             function ($x) {
                 return (int) $x;
             },
             $this->members
+        );
+        $this->nonMembers = array_map(
+            function ($x) {
+                return (int) $x;
+            },
+            $this->nonMembers
         );
     }
 
@@ -328,7 +345,7 @@ class MyRadio_List extends ServiceAPI
     /**
      * Return all mailing lists.
      *
-     * @return MyRadio_User[]
+     * @return MyRadio_List[]
      */
     public static function getAllLists()
     {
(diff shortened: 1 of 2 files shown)


src/Classes/ServiceAPI/MyRadio_List.php
```diff
@@ -67,6 +67,13 @@ class MyRadio_List extends ServiceAPI
     private $members = [];
 
     /**
+     * This is the set of members have opted out of this 'optin' list.
+     *
+     * @var int[]
+     */
+    private $nonMembers = [];
```
LordAro: nonMembers doesn't seem like a very good name for this. Maybe optedOutMembers ? Actually, members, and the SQL it comes from, seems fairly confusing as well. The "mail_subscription::member_id" table/col appears to list the members that have opted out of the mailing list? Very odd

LloydW93: For opt-in lists, the table represents members who are opted-in. For autogenerated etc lists, it represents members who are opted *out*.

LordAro:
Yeah, this *definitely* needs to be made clearer

src/Classes/ServiceAPI/MyRadio_List.php
```diff
@@ -345,6 +368,27 @@ public static function getAllLists()
         return $lists;
     }
 
+    /**
+     * Return mailing lists a user has access to.
+     *
+     * @return MyRadio_List[]
+     */
+    public static function getUserLists($memberid = -1)
```
LordAro: memberid undocumented

src/Classes/ServiceAPI/MyRadio_List.php
```diff
@@ -345,6 +368,27 @@ public static function getAllLists()
...
+    {
+        $member = MyRadio_User::getInstance($memberid)->getID();
```
LordAro: member, which is actually a memberid, which is the same as the function parameter?

LordAro: what happens when memberid=-1 is passed in, as per default (why is there a default?)

src/Classes/ServiceAPI/MyRadio_List.php
```diff
@@ -331,11 +349,16 @@ public static function getByName($str)
     /**
      * Return all mailing lists.
      *
-     * @return MyRadio_User[]
+     * @return MyRadio_List[]
      */
-    public static function getAllLists()
+    public static function getAllLists($current = "")
```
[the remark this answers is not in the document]
mstratford: well, I guess the idea was to allow 3 states, all, current and only history.

src/Classes/ServiceAPI/MyRadio_List.php
mstratford: I've changed this and added nicer comment. Hopefully this is better now.

src/Classes/ServiceAPI/MyRadio_List.php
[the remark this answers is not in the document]
mstratford: I've changed it to a default of Null, if that helps...

src/Classes/ServiceAPI/MyRadio_List.php
mstratford: Given my changes to getAllLists, I'm not actually going to use this. I'm wondering if we should just get rid of it.

PR #597, opened 2016-11-16
Make some things static for the quotes api

sdhand:


sdhand:
I have done, they are. (To my knowledge the only place that uses them is the quotes board page itself).

PR #600, opened 2016-11-30
Removed open on start feature of the SIS selector.

mstratford:
Closes UniversityRadioYork/Issues#63

### DIFF
--- a/src/Public/js/sis.selector.js
+++ b/src/Public/js/sis.selector.js
@@ -49,10 +49,6 @@ var Selector = function () {
                 return;
             }
 
-            if (lastTime === 0) {
-                this.show();
-            }
-
             lastTime = parseInt(data['lastmod']);
             // When called bet selectStudio, this isn't what I think it is
             // @todo, see if that can be bound nicer


LordAro:
@LloydW93 any ideas why this was there in the first place?

LloydW93:
Just a design decision between probably me/Andy at the time.

The Studio Selector used to be (still is?) limited to
- People with The Power
- People logged in from Studio 1

so it seemed reasonable for it to be there and open to be visible.

AJDurant:
Or a copy paste mistake in making those panels.

I don't think the location thing is working properly, but it might be.

mstratford:
@AJDurant Well, just tried with a non-comp member from my room, didn't show SIS selector or webcam, so I assume it works at least somewhat.

AJDurant:
The permissions part works, not the location overriding said permissions
IIRC

LordAro:
I mean the bit that opens it when `lastTime === 0` (which i'm not even sure of the meaning of, semantically)

(Not how it's open by default anyway)

AJDurant:
Huh, I hadn't actually looked at it, only the emails... So lastTime is
updated each time a response is received on the long poll that ensures data
is kept in order. It would only ever be 0 on initialisation, before
receiving data, which would open it, but allow it to be closed. Not sure
why it was done like that.

mstratford:
Am I correct in saying that this pr is ready for merge?

PR #602, opened 2016-12-22
Added a Bio: label to profile

mstratford:
My first pedantic dabble into MyRadio :)


LordAro:
Changes themselves look fine, but why the change at all? (Not got an existing opinion either way)

mstratford:
Just because it felt like it was missing, consistency and all that jazz.

Issue #592, opened 2016-10-28
Add member id to any drop-down that lists members


ChrisTheBaron:
This will eliminate issues like https://github.com/UniversityRadioYork/Issues/issues/56 and https://github.com/UniversityRadioYork/MyRadio/issues/588 where people are selecting the wrong users.

So the drop-downs would look like:
- Testy McTesterson (#123123)
- Testy McTesterson (#34564)

And people would be able to tell the difference (in theory)

LloydW93:
I disagree with this on the grounds of it being pretty poor UX - there must be something better we can do here.

e.g. 
- Add a property to the field type that limits it to current/paid/presenter
- Display 'eduroam' field when there's multiple
- Merge into one option in the drop down, then if selected open a modal with information about each
- Display profile photo
- "Sort" by relation to current user

ChrisTheBaron:
> Add a property to the field type that limits it to current/paid/presenter

Doesn't guarantee elimination of duplicates.

> Display 'eduroam' field when there's multiple

This is a possible solution, except non-students have eduroam as `null`. Also I would go for showing it against every user to keep it consistent.

> Merge into one option in the drop down, then if selected open a modal with information about each

IMO that's even worse UX, the user won't be expecting a modal window when selecting something from a drop-down.

> Display profile photo

We'd have to redo the entire UI to show them at any reasonable size, plus loading times.

> "Sort" by relation to current user

Again doesn't guarantee elimination of duplicates.

ChrisTheBaron:
Summons @LordAro & @Brookke for their thoughts

Brookke:
for the non eduroam users how do they login? Can we use usernames?

LordAro:
^ That's the question, I'm seeing 1359 members with null eduroam, and most of those are missing local_aliases as well, which would've been my second guess for what's used as a username.

But I agree, eduroam (or some arbitrary `username` thing) would be preferential to memberid

mstratford:
I can work on this if you want, maybe I can put eduroam, with ID for fallover?

LordAro:
Seems reasonable, maybe go eduroam -> local_alias? Since that's what they'll be using to login if they don't have an eduroam entry

mstratford:
Sure, but just to check, is there any issues with not being able to tell if it is eduroam or local_alias?

LordAro:
Not sure what you mean?

mstratford:
[...]

PR #607, opened 2017-01-07
Show eduroam/local_alias/memberID in members drop-downs.

mstratford:
Fixes: UniversityRadioYork/Issues#56, #592 

Looks like the issues I was having before was due to caching, the code is working as it should now (after ignoring it for ages). It chooses eduroam first, followed by localAlias, followed by memberID as least preferred.  👍

### DIFF
--- a/src/Public/js/myradio.form.js
+++ b/src/Public/js/myradio.form.js
@@ -37,29 +37,22 @@ window.MyRadioForm = {
      */
         var memberFields = $('fieldset.myradiofrm input.member-autocomplete:not(.tt-hint):not(.tt-input)');
         if (memberFields.length > 0) {
-            var memberLookup = new Bloodhound(
-                
+            var memberLookup = new Bloodhound(            
                 {
                     datumTokenizer: Bloodhound.tokenizers.whitespace,
                     queryTokenizer: Bloodhound.tokenizers.whitespace,
-                    limit: 50,
+                    limit: 25,
                     dupDetector: function (remote, local) {
                         return local.memberid == remote.memberid;
                     },
                     prefetch: {
                         url: myradio.makeURL('MyRadio', 'a-findmember', {term: null, limit: 500})
                     },
-                    remote: {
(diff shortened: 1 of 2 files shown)


LloydW93:
I know this conversation has been going on for basically forever now, so I'll try not to ask for all the new functionality, but how about not showing any of these additional fields if there's no duplicates?

Can't remember if I've asked this before, but is *any* MyRadio user able to see any other user's email today?

adfw:
A 'normal' user can't, committee positions generally can see emails

LloydW93:
Then it's worth being aware that we're making information more "public" than it was before by doing this. Ofc, technically anyone at UoY can sign up whenever they want, but *also* anyone at UoY can also just type the name into gmail, so...

mstratford:
Erm, well it comes directly from a SQL query, to a JSON string to the list you get on screen, without some really clever stuff, I can’t see how I could do that… (Bearing in mind my SQL knowledge isn’t exactly huge.)

LloydW93:
I can't see a way of doing it without a fair bit of work either (or at least, without trying to remember how typeahead works), so meh.

adfw:
I suppose on the whole, these member-select dropdowns are mostly restricted by permissions anyway

src/Public/js/myradio.form.js
```diff
@@ -78,7 +75,15 @@ window.MyRadioForm = {
                                     //Fix typeahead not showing after hiding
                                     //TODO: Report this @ https://github.com/twitter/typeahead.js/
                                     $('input:focus').parent().children('.tt-dropdown-menu').removeClass('hidden');
-                                    return '<p>' + i.fname + ' ' + i.sname + '</p>';
+                                    var identity
```
LordAro: don't like the lack of assignment here - `var identity = '';` ?

mstratford: Done & retested.

mstratford:
Fixed :) Using the lack of much TypeAhead/Bloodhound documentation, all I know is that it is possible to hide duplicates client side, but that doesn't count if the different details are provided through JSON first.

PR #610, opened 2017-01-08
Add the ability to set the output device, just like a real BAPS.

LloydW93:
I should probably add a description for this one!

This PR (along with a couple small Vagrant fixes), makes uses of the draft Audio Output Devices API, which has been available in Chrome for about a year now.

It enables you to configure a HTMLMediaElement's hardware output device, although with the caveat that the journey in for requesting access in the browser actually asks if you want to grant Microphone access, as the permission for Input/Output is the same.

By clicking on the channel name (e.g. "Channel 1") at the bottom of a channel, a new Channel Settings dialog appears, which currently only has this one new setting.

When this dialog is first opened, Chrome will ask the user for access to the microphone, and if granted the dropdown box shows an enumeration of available hardware output devices. These are numbered, and the mapping from number -> guid is persisted in local storage to prevent numbers flipping around from device connection/disconnection or browsers not having a deterministic API.

When the value of the input box is changed, the output of the channel is immediately changed, and this output mapping becomes stored in local storage too. Subsequent visits to Show Planner will reload the local storage configuration, re-request Microphone access if necessary, therefore persisting this configuration.

### DIFF
--- a/sample_configs/bootstrap.sh
+++ b/sample_configs/bootstrap.sh
@@ -70,4 +70,4 @@ service apache2 start
 
 # Somewhere to store audio uploads
 mkdir -p /music/records
-chown apache2:apache2 /music/records
+chown www-data:www-data /music/records
--- a/sample_configs/bootstrap.sh
+++ b/sample_configs/bootstrap.sh
@@ -17,7 +17,7 @@ apt-get install -y apache2 \
 	php-pear \
 	php5-memcached \
 	openssl \
-  libav-tools
+	libav-tools
 a2enmod ssl
 a2enmod rewrite
 service apache2 stop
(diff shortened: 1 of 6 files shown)


sample_configs/bootstrap.sh
```diff
@@ -16,7 +16,8 @@ apt-get install -y apache2 \
 	php5-dev \
 	php-pear \
 	php5-memcached \
-	openssl
+	openssl \
+  libav-tools
```
adfw: indentation :P

src/Public/js/nipsweb.channelconfig.js
```diff
@@ -0,0 +1,130 @@
...
+
+        var storedDevices = [];
+
+        try {
+            if (localStorage && localStorage.hasOwnProperty('nipsWebKnownDevices')) {
+                storedDevices = localStorage.nipsWebKnownDevices;
+            }
+        } catch (e) {console.error('Local Storage is being mean.', e);}
+
+        for (var i = 0; i < devices.length; i++) {
```
LordAro: foreach instead of for(var i... ) ?

same for j below

sample_configs/bootstrap.sh
```diff
@@ -66,3 +67,7 @@ su - postgres -c "cat /vagrant/sample_configs/postgres.sql | psql"
 
 # Start httpd back up
 service apache2 start
+
+# Somewhere to store audio uploads
+mkdir -p /music/records
```
LloydW93: Right now, that change just makes things work as they are. Editing the default should be its own issue/change.

src/Public/js/nipsweb.channelconfig.js
LloydW93: Sorry, I'm used to having to support oldIE etc, which doesn't have such fancy things. Might even get some ES2015 in here instead :P

LloydW93:
Re: Unsupported browsers, it either won't list devices, or when trying to switch will catch the failure and set back to default.

AJDurant:
Did you mean to put all of @mstratford's or in here?

LloydW93:
Not really, but it won't matter as long as his PR is merged first...

mstratford:
Can confirm it works in Studio 1 👍

Issue #613, opened 2017-01-09
Auto fill in MP3 Metadata on manual file upload


mstratford:
# **"with a big "please check this is all correct" notice"**

PR #614, opened 2017-01-11
Remove LastFm, implement auto ID3 tagging and improve UI of Track Uploader.

mstratford:
Fixes: #613 

**New features added**
- Nice bootstrap styling everywhere.
- Uploading to Central Music Library now auto-fills from ID3 tags, including if track has "Explicit" in title.

**Fixes**
- Bitrate/Stereo Errors when uploading are now displayed when not uploading to the central music library.
- When uploading to the Central Music Library, if a track fails the bitrate or stereo test, the user can no longer still submit the track to the library.
- For the time being (until get round to making multi-uploads work properly), only one Central Library Track can be uploaded and submitted at a time. Once a track is submitted, another can be uploaded. If a new track is uploaded before one is submitted, the old track will be shown as not submitted, instead of the blue alert box being left on the page in a non-working manor.
- The broken LastFM integration has been removed.

**Testing Performed So Far**

_Central Music Library:_

- Bitrate error: Locally & myradio-dev.
- Stereo error: Locally & myradio-dev.
- Uploading & Submitting: Local & myradio-dev
- Auto ID3 Track Listing: Local & myradio-dev

_Resources:_

- Bitrate error: myradio-dev only (not available locally).
- Stereo error: myradio-dev only.
- Just Uploading: myradio-dev only.
- Uploading & submitting: Not yet.
- Jingles - Expiry Dates: Not yet.

_My Beds/Jingles_

- Bitrate error: Locally & myradio-dev.
- Stereo error: Locally & myradio-dev.
- Uploading & submitting: myradio-dev only

**Known Bugs**

- (FIXED) Manual Track Listing presence check errors don't quite fit the form.
- (FIXED) Manual Jingle etc. name presence checks cause a MyRadio Exception (as before).

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Track.php
+++ b/src/Classes/ServiceAPI/MyRadio_Track.php
@@ -626,31 +626,29 @@ class MyRadio_Track extends ServiceAPI
         // File quality checks
         if ($fileInfo['audio']['bitrate'] < 192000) {
             return ['status' => 'FAIL', 'message' => 'Bitrate is below 192kbps', 'fileid' => $filename, 'bitrate' => $fileInfo['audio']['bitrate']];
-        } else if (strpos($fileInfo['audio']['channelmode'], 'stereo') === false) {
+        }
+        if (strpos($fileInfo['audio']['channelmode'], 'stereo') === false) {
             return ['status' => 'FAIL', 'message' => 'Item is not stereo', 'fileid' => $filename, 'channelmode' => $fileInfo['audio']['channelmode']];
-        } else {
-            $analysis['status'] = 'INFO';
-            $analysis['message'] = 'Currently editing track information for';
-            $analysis['submittable'] = True;
-            $analysis['fileid'] = $filename;
-            $analysis['analysis']['title'] = $fileInfo['comments_html']['title'];
-            $analysis['analysis']['artist'] = $fileInfo['comments_html']['artist'];
-            $analysis['analysis']['album'] = $fileInfo['comments_html']['album'];
+        }
+
+        $analysis['status'] = 'INFO';
+        $analysis['message'] = 'Currently editing track information for';
+        $analysis['submittable'] = True;
+        $analysis['fileid'] = $filename;
+        $analysis['analysis']['title'] = $fileInfo['comments_html']['title'];
+        $analysis['analysis']['artist'] = $fileInfo['comments_html']['artist'];
+        $analysis['analysis']['album'] = $fileInfo['comments_html']['album'];
             
            
-            //Remove total tracks in album from the track_number tag.
-            $trackNo = explode("/", $fileInfo['comments_html']['track_number'][0], 2)[0];
-            $analysis['analysis']['position'] = (string)$trackNo;
+        //Remove total tracks in album from the track_number tag.
+        $trackNo = explode("/", $fileInfo['comments_html']['track_number'][0], 2)[0];
+        $analysis['analysis']['position'] = (string)$trackNo;
 
-            $trackName = implode("", $fileInfo['comments_html']['title']);
+        $trackName = implode("", $fileInfo['comments_html']['title']);
            
-            if (stripos($trackName, 'explicit') == true) {
-                $analysis['analysis']['explicit'] = true;
-            } else {
-                $analysis['analysis']['explicit'] = false;
-            }
-            return $analysis;
-        }
+        $analysis['analysis']['explicit'] = !!stripos($trackName, 'explicit');
+
+        return $analysis;
     }
 
     /**
--- a/src/Classes/ServiceAPI/MyRadio_Track.php
+++ b/src/Classes/ServiceAPI/MyRadio_Track.php
@@ -636,9 +636,14 @@ class MyRadio_Track extends ServiceAPI
             $analysis['analysis']['title'] = $fileInfo['comments_html']['title'];
             $analysis['analysis']['artist'] = $fileInfo['comments_html']['artist'];
             $analysis['analysis']['album'] = $fileInfo['comments_html']['album'];
-            $analysis['analysis']['position'] = $fileInfo['comments_html']['track_number'][0];
+            
+           
+            //Remove total tracks in album from the track_number tag.
+            $trackNo = explode("/", $fileInfo['comments_html']['track_number'][0], 2)[0];
+            $analysis['analysis']['position'] = (string)$trackNo;
 
             $trackName = implode("", $fileInfo['comments_html']['title']);
+           
             if (stripos($trackName, 'explicit') == true) {
                 $analysis['analysis']['explicit'] = true;
             } else {
--- a/src/Classes/ServiceAPI/MyRadio_Track.php
+++ b/src/Classes/ServiceAPI/MyRadio_Track.php
@@ -625,36 +625,26 @@ class MyRadio_Track extends ServiceAPI
 
         // File quality checks
         if ($fileInfo['audio']['bitrate'] < 192000) {
-            return ['status' => 'FAIL', 'error' => 'Bitrate is below 192kbps.', 'fileid' => $filename, 'bitrate' => $fileInfo['audio']['bitrate']];
-        }
-        if (strpos($fileInfo['audio']['channelmode'], 'stereo') === false) {
-            return ['status' => 'FAIL', 'error' => 'Item is not stereo.', 'fileid' => $filename, 'channelmode' => $fileInfo['audio']['channelmode']];
-        }
-        if (strpos($fileInfo['audio']['channelmode'], 'stereo') === false) {
-            return ['status' => 'FAIL', 'error' => 'Item is not stereo.', 'fileid' => $filename, 'channelmode' => $fileInfo['audio']['channelmode']];
-        }
-
-        $analysis = self::identifyUploadedTrack(Config::$audio_upload_tmp_dir.'/'.$filename);
-        if (isset($analysis['status'])) {
-            //If song can't be found in Last FM'
-            $analysis['error'] = $analysis['error'] . ' Track metadata will be automatically filled-in below if embeded within the file.';
+            return ['status' => 'FAIL', 'message' => 'Bitrate is below 192kbps', 'fileid' => $filename, 'bitrate' => $fileInfo['audio']['bitrate']];
+        } else if (strpos($fileInfo['audio']['channelmode'], 'stereo') === false) {
+            return ['status' => 'FAIL', 'message' => 'Item is not stereo', 'fileid' => $filename, 'channelmode' => $fileInfo['audio']['channelmode']];
+        } else {
+            $analysis['status'] = 'INFO';
+            $analysis['message'] = 'Currently editing track information for';
+            $analysis['submittable'] = True;
             $analysis['fileid'] = $filename;
             $analysis['analysis']['title'] = $fileInfo['comments_html']['title'];
-            $analysis['analysis']['artist'] = $fileInfo['comments_html']['artist'][0];
+            $analysis['analysis']['artist'] = $fileInfo['comments_html']['artist'];
             $analysis['analysis']['album'] = $fileInfo['comments_html']['album'];
             $analysis['analysis']['position'] = $fileInfo['comments_html']['track_number'];
-            if (strpos($fileInfo['comments_html']['title'], 'explicit') == true) {
+
+            $trackName = implode("", $fileInfo['comments_html']['title']);
+            if (stripos($trackName, 'explicit') == true) {
                 $analysis['analysis']['explicit'] = true;
             } else {
                 $analysis['analysis']['explicit'] = false;
             }
             return $analysis;
-        } else {
-            return [
-                'fileid' => $filename,
-                'status' => 'OK',
-                'analysis' => $analysis,
-            ];
         }
     }
 
--- a/src/Classes/ServiceAPI/MyRadio_Track.php
+++ b/src/Classes/ServiceAPI/MyRadio_Track.php
@@ -620,7 +620,7 @@ class MyRadio_Track extends ServiceAPI
         $getID3 = new \getID3();
         $fileInfo = $getID3->analyze(Config::$audio_upload_tmp_dir.'/'.$filename);
         $getID3_lib = new \getID3_lib();
-        $getID3_lib->CopyTagsToComments($fileInfo);       
+        $getID3_lib->CopyTagsToComments($fileInfo);
 
         // File quality checks
         if ($fileInfo['audio']['bitrate'] < 192000) {
@@ -697,6 +697,7 @@ class MyRadio_Track extends ServiceAPI
             return $tracks;
         }
     }
+    
     /**
       * Pay special attention to the tri-state value of explicit. False and null are different things.
     */
--- a/src/Classes/ServiceAPI/MyRadio_Track.php
+++ b/src/Classes/ServiceAPI/MyRadio_Track.php
@@ -619,6 +619,8 @@ class MyRadio_Track extends ServiceAPI
 
         $getID3 = new \getID3();
         $fileInfo = $getID3->analyze(Config::$audio_upload_tmp_dir.'/'.$filename);
+        $getID3_lib = new \getID3_lib();
+        $getID3_lib->CopyTagsToComments($fileInfo);
         
 
         // File quality checks
@@ -637,11 +639,11 @@ class MyRadio_Track extends ServiceAPI
             //If song can't be found in Last FM'
             $analysis['error'] = $analysis['error'] . ' Track metadata will be automatically filled-in below if embeded within the file.';
             $analysis['fileid'] = $filename;
-            $analysis['analysis']['title'] = $fileInfo['tags_html']['id3v1']['title'];
-            $analysis['analysis']['artist'] = $fileInfo['tags_html']['id3v1']['artist'][0];
-            $analysis['analysis']['album'] = $fileInfo['tags_html']['id3v1']['album'];
-            $analysis['analysis']['position'] = $fileInfo['tags_html']['id3v2']['track_number'];
-            if (strpos($fileInfo['tags_html']['id3v1']['title'], 'explicit') !== false) {
+            $analysis['analysis']['title'] = $fileInfo['comments_html']['title'];
+            $analysis['analysis']['artist'] = $fileInfo['comments_html']['artist'][0];
+            $analysis['analysis']['album'] = $fileInfo['comments_html']['album'];
+            $analysis['analysis']['position'] = $fileInfo['comments_html']['track_number'];
+            if (strpos($fileInfo['comments_html']['title'], 'explicit') == true) {
                 $analysis['analysis']['explicit'] = true;
             } else {
                 $analysis['analysis']['explicit'] = false;
--- a/src/Classes/ServiceAPI/MyRadio_Track.php
+++ b/src/Classes/ServiceAPI/MyRadio_Track.php
@@ -697,7 +697,9 @@ class MyRadio_Track extends ServiceAPI
             return $tracks;
         }
     }
-
+    /**
+      * Pay special attention to the tri-state value of explicit. False and null are different things.
+    */
     public static function identifyAndStoreTrack($tmpid, $title, $artist, $album, $position, $explicit = null)
     {
         // We need to rollback if something goes wrong later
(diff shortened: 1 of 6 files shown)


src/Classes/ServiceAPI/MyRadio_Track.php
```diff
@@ -619,26 +619,37 @@ public static function cacheAndIdentifyUploadedTrack($tmp_path)
...
-        if (strpos($fileInfo['audio']['channelmode'], 'stereo') === false) {
-            return ['status' => 'FAIL', 'error' => 'Item is not stereo.', 'fileid' => $filename, 'channelmode' => $fileInfo['audio']['channelmode']];
-        }
-
-        $analysis = self::identifyUploadedTrack(Config::$audio_upload_tmp_dir.'/'.$filename);
-        if (isset($analysis['status'])) {
+            return ['status' => 'FAIL', 'message' => 'Bitrate is below 192kbps', 'fileid' => $filename, 'bitrate' => $fileInfo['audio']['bitrate']];
+        } else if (strpos($fileInfo['audio']['channelmode'], 'stereo') === false) {
+            return ['status' => 'FAIL', 'message' => 'Item is not stereo', 'fileid' => $filename, 'channelmode' => $fileInfo['audio']['channelmode']];
+        } else {
```
LordAro: Noticed this on an exceptionally brief look: Could change the if {} else if {} to separate if {} statements (since they just return from the function), then you could reduce the indentation by getting rid of the else {}

src/Classes/ServiceAPI/MyRadio_Track.php
```diff
@@ -619,26 +619,37 @@ public static function cacheAndIdentifyUploadedTrack($tmp_path)
...
+            
+           
+            //Remove total tracks in album from the track_number tag.
+            $trackNo = explode("/", $fileInfo['comments_html']['track_number'][0], 2)[0];
+            $analysis['analysis']['position'] = (string)$trackNo;
+
+            $trackName = implode("", $fileInfo['comments_html']['title']);
+           
+            if (stripos($trackName, 'explicit') == true) {
+                $analysis['analysis']['explicit'] = true;
```
LordAro: Some boolean logic required here?

LloydW93:
UI thought: Should we change the position of "Save using below metadata" (and the copy too) to *below* the form, so that the user has to actually read the metadata first?

LloydW93:
Probably not a new bug, but since we've been talking about it:
- Drag in a track
- Hit "Save..." button
- While it's transcoding, drag in a second track
- Saving bar turns red and displays an error after the saving message
- Red bar eventually changes to a tick, but is red

src/Classes/ServiceAPI/MyRadio_Track.php
mstratford: You mean get rid of the == true?

LordAro: I mean merge it all into one line. No need for `if (something) { foo = true; } else { foo = false;}` just do `foo = something;`

mstratford:
I agree with putting the submit at the bottom of the form, I'll give it a go :)

src/Classes/ServiceAPI/MyRadio_Track.php
mstratford: Oh, yeah 👍

LloydW93: To be right type-wise, do foo = !!something

mstratford:
OK, moved the button and fixed the red message :)

src/Public/js/nipsweb.managelibrary.js
```diff
@@ -199,22 +221,24 @@ var Library = function () {
...
                                 }
                             );
                         }
                     );
 
-                    result.append('<label for="centralupload-' + i + '">' + file.name + ':&nbsp;</label>')
-                    .append(select)
-                    .append(submit);
+                    result.append('<label for="centralupload-' + i + '">' + file.name + ' &nbsp;</label>');
+                    //.append(select)
```
AJDurant: Could be removed?

src/Public/js/nipsweb.managelibrary.js
```diff
@@ -235,49 +259,72 @@ var Library = function () {
...
                 uploadFinished: function (i, file, response, time) {
-                    $('#res-status').html('Uploaded ' + file.name);
+                    $('#res-status').html(icon_ok + 'Uploaded ' + file.name);
 
                     var result = $('<div class="alert"></div>');
                     if (response['status'] == 'FAIL') {
-                        //An error occurred
-                        result.addClass('alert-danger').append('<span class="error">' + response['error'] + '</span>');
+                        //An error occurred, probably bitrate is too low.
+                        result.addClass('alert-danger alert-dismissable').append(icon_error + icon_close + '<strong>' + file.name + ':</strong> <span class="error">' + response['error'] + '</span>');
```
AJDurant: Should this be response['message'] now?

src/Public/js/nipsweb.managelibrary.js
mstratford: Yup, sorry.

src/Public/js/nipsweb.managelibrary.js
mstratford: Will change :)

src/Classes/ServiceAPI/MyRadio_Track.php
```diff
@@ -619,27 +619,36 @@ public static function cacheAndIdentifyUploadedTrack($tmp_path)
 
         $getID3 = new \getID3();
         $fileInfo = $getID3->analyze(Config::$audio_upload_tmp_dir.'/'.$filename);
+        $getID3_lib = new \getID3_lib();
+        $getID3_lib->CopyTagsToComments($fileInfo);
```
LordAro: Is it necessary to assign this to a variable? IIRC you can do something like `(new \getID3_lib())->...` or maybe there's something even better - is it static, perhaps?

src/Classes/ServiceAPI/MyRadio_Track.php
```diff
@@ -619,27 +619,36 @@ public static function cacheAndIdentifyUploadedTrack($tmp_path)
...
+
```
LordAro: unnecessary newline? </pedant>

src/Classes/ServiceAPI/MyRadio_Track.php
```diff
@@ -619,27 +619,36 @@ public static function cacheAndIdentifyUploadedTrack($tmp_path)
...
-            ];
-        }
+        $analysis['status'] = 'INFO';
+        $analysis['message'] = 'Currently editing track information for';
+        $analysis['submittable'] = True;
+        $analysis['fileid'] = $filename;
+        $analysis['analysis']['title'] = $fileInfo['comments_html']['title'];
+        $analysis['analysis']['artist'] = $fileInfo['comments_html']['artist'];
+        $analysis['analysis']['album'] = $fileInfo['comments_html']['album'];
+
```
LordAro: extra newline?

src/Classes/ServiceAPI/MyRadio_Track.php
```diff
@@ -648,6 +657,8 @@ public static function cacheAndIdentifyUploadedTrack($tmp_path)
      * !This method requires the external lastfm-fpclient application to be installed on the server. A FreeBSD build
      * with URY's API key and support for -json can be found in the fpclient.git URY Git repository.
      *
+     ***** Since LastFM was removed from the central track uploader, this code MAY not be used anymore.
```
LordAro: If the code is dead, why not remove it?

src/Classes/ServiceAPI/MyRadio_Track.php
```diff
@@ -690,10 +701,6 @@ public static function identifyUploadedTrack($path)
         }
     }
 
-
-    /**
-     * Pay special attention to the tri-state value of explicit. False and null are different things.
```
LordAro: I don't think this should be getting removed...

src/Controllers/NIPSWeb/confirm_central_upload.php
```diff
@@ -11,7 +11,7 @@
     $_REQUEST['artist'],
     $_REQUEST['album'],
     $_REQUEST['position'],
-    $_REQUEST['explicit'] ? true : null
+    filter_var($_REQUEST['explicit'], FILTER_VALIDATE_BOOLEAN) ? true : null
```
LordAro: From the docs for filter_var: "Returns the filtered data, or FALSE if the filter fails."

Seems to me that this won't be able to differentiate between `false` and some not-boolean value. Sure this is the right thing to do?

src/Public/css/planner.css
```diff
@@ -58,29 +58,68 @@ body {
...
+  margin-top: 6px;
+}
+.result-container .alert-dismissable {
+    padding-right: 35px;
+}
+/* Spin while saving tracks */
+.gly-spin {
+  -webkit-animation: spin 2s infinite linear;
+  -moz-animation: spin 2s infinite linear;
+  -o-animation: spin 2s infinite linear;
```
LordAro: Do we still need the css prefixes? Hasn't CSS3 all been nice and standardised for ages now? Applies to the bits below this as well

src/Public/css/planner.css
```diff
@@ -58,29 +58,68 @@ body {
...
-  top: calc(100% - 70px);
-  right: 1em;
-  text-align: right;
+  margin: 5px 0;
+}
+.result-container .alert label {
+  margin-top: 6px;
+}
+.result-container .alert-dismissable {
+    padding-right: 35px;
```
LordAro: inconsistent indent - should be 2 spaces

src/Public/js/nipsweb.managelibrary.js
```diff
@@ -4,8 +4,13 @@ var Library = function () {
...
     var allowed_all = ['audio/mpeg3', 'audio/x-mpeg-3', 'audio/mpeg', 'audio/x-mpeg',
             'audio/mp3', 'audio/x-mp3', 'audio/mpg', 'audio/mpg3', 'audio/mpegaudio', 'audio/wav', 'audio/x-wav',
-            'audio/mp4a-latm', 'audio/mp4', 'audio/aac']
+            'audio/mp4a-latm', 'audio/mp4', 'audio/aac'];
 
+    var icon_error = '<div class="glyphicon glyphicon-exclamation-sign"></div>&nbsp;';
+    var icon_ok = '<div class="glyphicon glyphicon-ok"></div>&nbsp;';
+    var icon_ok = '<div class="glyphicon glyphicon-ok"></div>&nbsp;';
+    var icon_loading = '<div class="glyphicon glyphicon-refresh gly-spin"></div>&nbsp;';
+    var icon_close = '<a href="#" class="close" data-dismiss="alert" aria-label="close">&times;</a>';
```
LordAro: "constants", in my book at least, should be in ALL_CAPS. not sure if JS has any sort of "const" directive...

src/Public/js/nipsweb.managelibrary.js
```diff
@@ -67,52 +72,63 @@ var Library = function () {
...
-                            // If the div exists, then the user has permission to upload a track
-                            // manually, so display the div and set manual_track to true.
-                            $(manual_div).show();
-                            manual_track = true;
-                        }
-                    }
-
+                            if (response['status'] !== 'OK' && response['submittable'] == true ) {
+                                
+                                    // If the div exists, then the user has permission to upload a track
```
LordAro: formatting (among others) JS is also 2-spaces

I know several existing JS files don't follow that, but that's for another time..

src/Public/js/nipsweb.managelibrary.js
```diff
@@ -67,52 +72,63 @@ var Library = function () {
...
                         function () {
                             if ($('#central-status').html() == status) {
-                                $('#central-status').html('Ready');
+                                $('#central-status').html(icon_ok + 'Ready');
                             }
                         },
                         5000
                     )
 
                     var manual_track = false;
```
LordAro: the variable name `manual_track` doesn't quite fit anymore, maybe `submittable_track`?

src/Public/js/nipsweb.managelibrary.js
```diff
@@ -122,34 +138,35 @@ var Library = function () {
...
-                        .attr('name', response.fileid).attr('id', 'centralupload-' + i);
-                        $.each(
-                            response.analysis,
-                            function (
-                                key,
-                                value
-                            ) {
-                                select.append('<option value="' + value.title + ':-:' + value.artist + '">' + value.title + ' by ' + value.artist + '</option>');
+
+                    if (manual_track) {
```
LordAro: Seems to be quite a lot of code that only does stuff if manual_track is set. Can some early returns be used to reduce the indentation?

src/Public/js/nipsweb.managelibrary.js
```diff
@@ -122,34 +138,35 @@ var Library = function () {
...
+                        if (response.analysis) {
+                            //Otherwise, if we got an analysis from the ID3 tags, prefill the textboxes.
+                            function decodeHtml(html) {
```
LordAro: function here is a bit odd. Maybe a lambda instead?

src/Public/js/nipsweb.managelibrary.js
```diff
@@ -199,22 +221,23 @@ var Library = function () {
...
                                 }
                             );
                         }
                     );
 
-                    result.append('<label for="centralupload-' + i + '">' + file.name + ':&nbsp;</label>')
-                    .append(select)
-                    .append(submit);
+                    result.append('<label for="centralupload-' + i + '">' + file.name + ' &nbsp;</label>');
+                    if (manual_track == true) {
```
LordAro: `== true` :(

src/Public/js/nipsweb.managelibrary.js
```diff
@@ -235,49 +258,72 @@ var Library = function () {
...
+                                        result.find('input.title').prop('disabled', true); 
+                                        setTimeout(
+                                            function () {
+                                                result.find('input.title').prop('disabled', false); 
+                                                result.find('input.title').focus();
+                                            },
+                                            500
+                                        )
+                                        
+                                        return;
```
LordAro: some odd amount of spacing for this section, squish it up a bit?

src/Templates/NIPSWeb/manage_library_manual.twig
```diff
@@ -8,27 +8,47 @@
...
+                  </div>
+                </form>
+              </div>
           </div>
 
         </div>
         <div id="managed-container" class="res-container hidden">Managed playlists are controlled using <a href="/itones/" target="_blank">Campus Playout Manager</a>. Entering them there also adds them to the wonderful Campus Jukebox</div>
         <div id="res-container" class="res-container hidden">
-          <p>You can upload track to this resource library by dragging them into the box below. Tracks uploaded must not return a match against our music analysis software.</p>
+          <p><br>You can upload track to this resource library by dragging them into the box below. Tracks uploaded must not return a match against our music analysis software.</p>
```
LordAro: maybe move the br to after the div.managed-container?

src/Classes/ServiceAPI/MyRadio_Track.php
mstratford: Internal Server error when you try that example :(

src/Classes/ServiceAPI/MyRadio_Track.php
mstratford: It's used by MyRadio_FingerPrinterDaemon.php, don't want to mess anything up if that's still running somewhere...

src/Controllers/NIPSWeb/confirm_central_upload.php
mstratford: @LloydW93 Not sure on this one?

src/Public/js/nipsweb.managelibrary.js
mstratford: Erm, well this is apparently the best method I could find on Google for doing it this way round? Converting & to &amp; etc is easy, but apparently the other way is hard...

src/Controllers/NIPSWeb/confirm_central_upload.php
mstratford: Well, I just tested that Explicit checking with this is working properly, so is there really a big deal?

src/Classes/ServiceAPI/MyRadio_Track.php
AJDurant: Remove the fingerprinter daemon too? Iirc that scanned tracks to check if it seemed like the data we had was correct for them. But the lastfm fingerprinter is not coming back

src/Controllers/NIPSWeb/confirm_central_upload.php
AJDurant: Was it broken without filter_var? I've never seen it used before

mstratford:
I assume then that it definitely doesn’t run anymore? If so, I’ll have a look.

mstratford:
Whatever was there directly before it seemed to work. However LloydW didn’t seem to think it did.

src/Controllers/NIPSWeb/confirm_central_upload.php
mstratford: Whatever was there directly before it seemed to work. However LloydW didn’t seem to think it did.

src/Classes/ServiceAPI/MyRadio_Track.php
mstratford: If that definitely doesn’t run anymore, I’ll have a look.

src/Public/js/nipsweb.managelibrary.js
AJDurant: In ES6 you have const and let instead of var, well var is still there but you're not supposed to use it.

Looks like const has decent browser support.

mstratford: Yup, I tried it and it worked with const, so I've already updated it :)

src/Classes/ServiceAPI/MyRadio_Track.php
```diff
@@ -620,8 +620,7 @@ public static function cacheAndIdentifyUploadedTrack($tmp_path)
         $getID3 = new \getID3();
         $fileInfo = $getID3->analyze(Config::$audio_upload_tmp_dir.'/'.$filename);
         $getID3_lib = new \getID3_lib();
-        $getID3_lib->CopyTagsToComments($fileInfo);
-        
+        $getID3_lib->CopyTagsToComments($fileInfo);
```
LordAro: Trailing whitespace :>

src/Classes/ServiceAPI/MyRadio_Track.php
```diff
@@ -700,7 +697,9 @@ public static function identifyUploadedTrack($path)
             return $tracks;
         }
     }
-
```
LordAro: missing newline between functions :>

src/Classes/ServiceAPI/MyRadio_Track.php
AJDurant: You should probably set your editor up to remove trailing whitespace, and to add newlines on the ends of files

src/Classes/ServiceAPI/MyRadio_Track.php
mstratford: Done

src/Classes/ServiceAPI/MyRadio_Track.php
mstratford: Done :)

mstratford: Fixed

src/Classes/ServiceAPI/MyRadio_Track.php
mstratford: Fixed

src/Controllers/NIPSWeb/confirm_central_upload.php
LloydW93: What was there directly before it did not work, because "false" the string evaluates to true.

src/Public/js/nipsweb.managelibrary.js
```diff
@@ -332,11 +373,9 @@ var Library = function () {
...
 
-
-
 $(document).ready(
     function () {
         library = Library();
         library.initialise();
     }
-);
+);
```
LloydW93: :O

mstratford: fixed 👍 I've also set my code editor to do add these automatically :)

PR #615, opened 2017-01-12
Made messages look more stylish, and (re)added stars to unread icons.

### LINKED ISSUE #372: [SIS] Unread messages have no labeling
previously unread messages were marked as such with a star using css on the 'unread' class

mstratford:
Fixes #372 

Also makes the popups more pretty and adds a few tooltips 👍 

**Testing Completed**

I used my own local copies of sis.css and sis.messages.js to override on live SIS, so should be ready to go 👍


mstratford:
http://imgur.com/a/W7xH4 for a preview :)

Brookke:
hmm i like although not sure starts are helpful as they usually indicate a favourite in most UI's ( I know they previously were starts though)  a circle might be more appropriate.

mstratford:
I'll eye-drop the colour of the iOS/MacOS mail unread dot and copy it 👍

Brookke:
haha doesn't have to be that exact ;)

AJDurant:
You could potentially make the preview text better, sometimes on long
messages it's a bit weird. Something like a fixed size trim with an
ellipsis?

mstratford:
Was thinking about that, i’ll get on it now :)

mstratford:
OK, I've done all of that :) Ready for merge!

src/Public/css/sis.css
```diff
@@ -45,9 +45,22 @@
...
+.td-msgitem.unread td:first-child {
+  visibility: visible;
+}
+.td-msgitem td:nth-child(2) {
+  width: 15px;
+  padding-right: 35px;
+}
+.td-msgitem .unread-dot {
+  border-radius: 100%;
+  height:14px;
```
Brookke: might be worth using em here then they scale with the font size

mstratford: Font size doesn't change in show planner though by the looks of it?

Brookke: True, but people might be running their browsers with a different font size.

mstratford: done :)

src/Public/js/sis.messages.js
```diff
@@ -86,7 +86,14 @@ var Messages = function () {
 
                 locationTd.innerHTML = icon_location;
                 unreadTd.innerHTML = icon_unread;
-                titleTd.innerHTML = data[i]['title'];
+
+                var longTitle = data[i]['title'];
+                var trimmedTitle = longTitle.substring(0, 120);
+                if (longTitle != trimmedTitle) {
+                    trimmedTitle = trimmedTitle + '... <a>Read more</a>';
```
AJDurant: If someone clicks on read more, does it still work, or does it's anchor tag with no destination override to do nothing?

mstratford: Just tried it, it doesn't have any effect, so still opens the popup.

AJDurant: Excellent

PR #617, opened 2017-01-12
Add a track search API call

LloydW93:
* Fix only one parameter having a type documented in the API.
* Fix putting single-parameter parameters into the path if they're arrays.
* Add a MyRadio_Track search method that can be used over the API.
* Fix findByOptions, because it turns out it's been broken since we moved the SQL out of the MyRadio_Track constructor and into getInstance.
* Tell Swagger not to expose findByOptions because its signature doesn't make sense for the API.
* Suck at splitting this into multiple commits

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Swagger.php
+++ b/src/Classes/ServiceAPI/MyRadio_Swagger.php
@@ -200,7 +200,7 @@ class MyRadio_Swagger
                 $lines[] = '';
             } elseif (substr($line, 0, 1) === '@') {
                 $key = preg_replace('/^\@([a-zA-Z]+)(.*)$/', '$1', $line);
-                $keys[] = [ 'type' => $key, 'data' => trim(preg_replace('/^\@([a-zA-Z]+) (.*)$/', '$2', $line))];
+                $keys[] = ['type' => $key, 'data' => trim(preg_replace('/^\@([a-zA-Z]+) (.*)$/', '$2', $line))];
             } else {
                 $lines[sizeof($lines) - 1] .= $line.' ';
             }
--- a/src/Classes/ServiceAPI/MyRadio_Swagger.php
+++ b/src/Classes/ServiceAPI/MyRadio_Swagger.php
@@ -200,7 +200,7 @@ class MyRadio_Swagger
                 $lines[] = '';
             } elseif (substr($line, 0, 1) === '@') {
                 $key = preg_replace('/^\@([a-zA-Z]+)(.*)$/', '$1', $line);
-                $keys[$key][] = trim(preg_replace('/^\@([a-zA-Z]+) (.*)$/', '$2', $line));
+                $keys[] = [ 'type' => $key, 'data' => trim(preg_replace('/^\@([a-zA-Z]+) (.*)$/', '$2', $line))];
             } else {
                 $lines[sizeof($lines) - 1] .= $line.' ';
             }
@@ -221,8 +221,8 @@ class MyRadio_Swagger
         //Now parse for docblock things
         $params = [];
         $return_type = 'Set';
-        foreach ($doc['keys'] as $key => $values) {
-            switch ($key) {
+        foreach ($doc['keys'] as $key) {
+            switch ($key['type']) {
                 //Deal with $params
                 case 'param':
                     /*
@@ -231,7 +231,7 @@ class MyRadio_Swagger
                      * info[2] should be parameter name
                      * info[3] should be the description
                      */
-                    $info = explode(' ', $values[0], 4);
+                    $info = explode(' ', $key['data'][0], 4);
                     $arg = str_replace('$', '', $info[2]); //Strip the $ from variable name
                     $params[$arg] = ['type' => $info[1], 'description' => empty($info[3]) ?: $info[3]];
                     break;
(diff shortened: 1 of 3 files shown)


src/Classes/ServiceAPI/MyRadio_Swagger.php
```diff
@@ -200,7 +200,7 @@ public static function parseDoc($doc)
                 $lines[] = '';
             } elseif (substr($line, 0, 1) === '@') {
                 $key = preg_replace('/^\@([a-zA-Z]+)(.*)$/', '$1', $line);
-                $keys[$key][] = trim(preg_replace('/^\@([a-zA-Z]+) (.*)$/', '$2', $line));
+                $keys[] = [ 'type' => $key, 'data' => trim(preg_replace('/^\@([a-zA-Z]+) (.*)$/', '$2', $line))];
```
LordAro: extra space ;)

src/Classes/ServiceAPI/MyRadio_Swagger2.php
```diff
@@ -348,6 +370,44 @@ private static function getTags($apis)
...
 
+    private static function getMethodOpType($method)
+    {
+        $name = $method->getName();
+
+        //Note the ordering is important - create is static!
+        if (
+            CoreUtils::startsWith($name, 'create') ||
+            CoreUtils::startsWith($name, 'add')
+        ) {
```
LordAro: Don't think I'm a fan of this overly split style of if statements. Put it back on one line? I don't think it's that long anyway, and we've never overly cared about line length before...

src/Classes/ServiceAPI/MyRadio_Swagger2.php
```diff
@@ -348,6 +370,44 @@ private static function getTags($apis)
...
+            return 'post';
+        }
+
+        if (
+            $name === 'toDataSource' ||
+            CoreUtils::startsWith($name, 'get') ||
+            CoreUtils::startsWith($name, 'is') ||
+            $method->isStatic()
+        ) {
```
LordAro: split if

src/Classes/ServiceAPI/MyRadio_Swagger2.php
```diff
@@ -348,6 +370,44 @@ private static function getTags($apis)
...
+            return 'get';
+        }
+
+        return 'put';
+    }
+
+    private static function getMethodPublicName($method) {
+        $name = $method->getName();
+
+        if ($name === 'toDataSource' || $name === 'create') {
```
LordAro: add findByOptions here instead, if it's never meant to be exposed? Or alternatively, add `@swagger ignore` directives to toDataSource & create functions

src/Classes/ServiceAPI/MyRadio_Track.php
```diff
@@ -438,16 +438,73 @@ private static function findByNameArtist($title, $artist, $limit, $digitised = f
...
+    public static function search(
+        $title = null,
+        $artist = null,
+        $recordid = null,
+        $digitised = true,
+        $clean = null,
+        $precise = false,
+        $page = 1,
+        $sort = null
+    ) {
```
LordAro: split if^H^Hfunction declaration. this one will be a bit long, but I don't think it'll be too awful. No worse than MyRadio_User::create anyway

AJDurant:
@LordAro as in you don't like things being on more than one line? It makes
it a lot more readable

src/Classes/ServiceAPI/MyRadio_Track.php
LloydW93: MyRadio_User::create is formatted the same way.

src/Classes/ServiceAPI/MyRadio_Swagger2.php
LloydW93: A path of '' doesn't mean ignore, it means no extra path.

POST /api/v2/user/ -> MyRadio_User::create
GET /api/v2/user/1 -> MyRadio_User::toDataSource

src/Classes/ServiceAPI/MyRadio_Swagger.php
LloydW93: I'd say missing, actually, but yes, one of those.

src/Classes/ServiceAPI/MyRadio_Swagger2.php
LloydW93: Except the build will fail if there's more than 120 chars per line. I suspect this one would still be under, but I felt it made sense for both ifs in this method to be consistent.

src/Classes/ServiceAPI/MyRadio_Swagger2.php
LloydW93: See above.

src/Classes/ServiceAPI/MyRadio_Swagger2.php
LordAro: True, but I think it looks better if ifs are just split over 2 lines (with an extra indent for the additional line), rather than spread over several.

I'll concede the function call though

AJDurant: The codestyle is PSR4, this is PSR4

PR #618, opened 2017-01-13
Implement a Library Search page.

mstratford:
It didn't do anything, so I made it do something.


src/Controllers/Library/search.php
```diff
@@ -1,6 +1,25 @@
...
+    );
+}
+
+CoreUtils::getTemplateObject()->setTemplate('table.twig')
+    ->addVariable('tablescript', 'myury.library.search')
+    ->addVariable('tabledata', CoreUtils::dataSourceParser($tracks))
+    ->addVariable('title', 'Search Library')
+    ->addVariable(
+        'text',
+        'Here you can search for tracks in the Central Music Library. <form action="#" method="get" class="form-inline"><input type="text" name="title" size="30" class="form-control" placeholder="Search by Track" /><input type="text" name="artist" size="30" class="form-control" placeholder="Search by Artist" /><button class="btn btn-primary" type="submit">Submit</button></form>'
```
mstratford: I know, I shouldn't have done a PRQ really.

AJDurant:
Why did you close it. What you've done is good, it should just be using MyRadio forms instead of HTML inside text

src/Controllers/Library/search.php
```diff
@@ -1,6 +1,54 @@
...
+    new MyRadioFormField('title', MyRadioFormField::TYPE_TEXT, ['required' => false, 'label' => 'Title', 'placeholder' => 'Filter by track title...'])
+)->addField(
+    new MyRadioFormField('artist', MyRadioFormField::TYPE_TEXT, ['required' => false, 'label' => 'Artist', 'placeholder' => 'Filter by artist name...'])
+);
+        
+if ($_SERVER['REQUEST_METHOD'] === 'POST') {
+    //Submitted
+    $data = $form->readValues();
+
+    if (isset($data['title']) || isset($data['title'])) {
```
AJDurant: I think you mean artist for the second

src/Controllers/Library/search.php
```diff
@@ -1,6 +1,54 @@
...
 use \MyRadio\MyRadio\CoreUtils;
+use \MyRadio\MyRadio\URLUtils;
+use \MyRadio\ServiceAPI\MyRadio_Track;
+use \MyRadio\MyRadio\MyRadioForm;
+use \MyRadio\MyRadio\MyRadioFormField;
+
+$form = new MyRadioForm(
+    'lib_search',
+    'Library',
+    'search',
```
AJDurant: It might be better to use the module and controller variable for this instead of strings, idk. Some forms do.

src/Controllers/Library/search.php
```diff
@@ -1,6 +1,54 @@
...
+if ($_SERVER['REQUEST_METHOD'] === 'POST') {
+    //Submitted
+    $data = $form->readValues();
+
+    if (isset($data['title']) || isset($data['title'])) {
+        $tracks = MyRadio_Track::findByOptions(
+            [
+                    'title' => isset($data['title']) ? $data['title'] : '',
+                    'artist' => isset($data['artist']) ? $data['artist'] : '',
+                    'limit' => 30
```
AJDurant: What happens if there are more than 30? Can you page through?

src/Controllers/Library/search.php
mstratford: cheers :)

src/Controllers/Library/search.php
mstratford: Well, I thought, since the tables paginate automatically by default at 10 per page, by the time you've read 3 pages and not found it, you would probably improve your search entry to narrow it down. I just did this because that's what stuff like Show Planner does when you search for a track.

src/Controllers/Library/search.php
mstratford: Just looked at those, like this: https://github.com/UniversityRadioYork/MyRadio/blob/master/src/Classes/MyRadio/MyRadio_Availability.php#L366

It seems like you would only use that if it were inside a function? Or am I misunderstanding you?

src/Controllers/Library/search.php
mstratford: I can remove this if you want, just thought it may be good to limit incase a search pulls up a lot of songs...

src/Controllers/Library/search.php
AJDurant: They are set by the root initialise on the request. The reason for them is to tell the form where to submit. It probably makes most sense to leave them as is on second thought

src/Controllers/Library/search.php
AJDurant: It does make sense, but then if you are searching the library you probably want to search the whole library. I don't see much reason to limit it, other than you could get a *lot* of data back with a silly search. The standard way to do that would be to only fetch a page of results at a time, but we don't easily support that in MyRadio.

I think seeing as this is the place to search the library it might as well not be limited. Others may dispute this though.

Maybe leave it for now, it's easy enough to remove later.

mstratford:
Erm, can you possibly have a quick look into why, now when using sending the table data through the myradio forms template we seem to have lost TRUE or FALSE from the clean and digitised columns? I can't figure it out myself...

AJDurant:
Ah I wondered if it would work or not, but forgot. In the js that does the the table data, you might need to specify that it is boolean

mstratford:
It did work properly (showed true or false) when I was using the normal table template method at the start of the PRQ, but somehow now it only shows a 1 in the clean when the DB clean value is set to 'u' (unknown?) I have no idea how that has happened?

AJDurant:
Not sure, can't really tell from my phone. Perhaps make a new issue on it to fix later.

mstratford:
If I change the code slightly just to hide the column, can you merge that too?

PR #619, opened 2017-01-13
Made the webcam archive page actually display a page.

mstratford:
Just thought I would make this page actually be a page, y'know.


LordAro:
But what if i want to know what the current unix epoch is equal to?!

mstratford:
Haha ;) Maybe i'll get around to seeing if I can actually build fully, but it would be nice to get a user friendly page :)

LordAro:
But more seriously, not a fan of the commented out code. This is what we have version control systems for ;)

mstratford:
Sorry, there you go :)

PR #620, opened 2017-01-14
Removed columns that don't format nicely.

mstratford:


AJDurant:
Could you not make them format nicely?

mstratford:
I could, but at least the clean column having a 1 doesn't mean that it's clean... It's broken somehow...

mstratford:
I'd kind of like to merge this soon, since otherwise it's lying to you ;P

mstratford:
The album was output as an Array, not sure how to make it display nicely? I couldn't work out what it was an array of, but it looks like your (current live, not PR one) edit form also stays empty.

AJDurant:
Hmm, interesting

PR #623, opened 2017-01-14
Fixes non-defined input notices in MyRadio_Track::findByOptions.

### LINKED ISSUE #621: MyRadio_Track::findByOptions Not defined properly.
See: http://imgur.com/a/hvgIB

Looks like it's not doing the right kind of check to see if these parameters were entered. Spotted on myradio-dev on master branch on my new search page.

mstratford:
Fixes #621 

May not be efficient, but it got rid of the notices 👍  It just makes doubly sure that the values are set before the SQL query 👍 You'll want to merge this before updating MyRadio live.

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Track.php
+++ b/src/Classes/ServiceAPI/MyRadio_Track.php
@@ -206,7 +206,7 @@ class MyRadio_Track extends ServiceAPI
                 ]
             );
     }
-    
+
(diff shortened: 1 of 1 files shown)


LordAro:
Assigned @LloydW93 as it's his code.

(Pls be better at commit messages, it's important that we can read the history "at a glance)

PR #626, opened 2017-01-15
Use travis to test the API, version 2

LordAro:
Adds the following basic capabilities:
* Stand up MyRadio in a Travis test environment, with working API
* Use of jasmine-node and frisby to run API calls
* API available on :80 in Vagrant, to enable local running of API tests
* Added xdebug to Vagrant (cause a substantial performance hit - may be worth disabling by default)
* RFTM now provides documentation for create APIs, where specified using json-schema

Fixes:
* base.sql had a missing sequence causing the API not to work correctly on new installs
* Nested getCredits calls could fail via the API


LordAro:
Also, why on earth is swagger-ui.js 26000 lines long?

LloydW93:
This probably breaks UI pages for creating members.

src/Classes/ServiceAPI/MyRadio_Show.php
```diff
@@ -247,7 +247,7 @@ public static function create($params = []) {
         if (!empty($_SESSION['memberid'])) {
             $creator = $_SESSION['memberid'];
         } else {
-            $creator = $params['credits'][0]['memberid'];
+            $creator = $params['credits']['memberid'][0];
```
AJDurant: Is it a memberid array rather than an array of credits with memberids then?

AJDurant: Is it a memberid array rather than an array of credits with memberid keys?

src/Classes/ServiceAPI/MyRadio_Quote.php
```diff
@@ -253,7 +253,7 @@ public function setSource($source)
     /**
      * Sets this quote's date.
      *
-     * @param int|string $date The date, as a UNIX timestamp or date string.
+     * @param int $date The date, as a UNIX timestamp or date string.
```
AJDurant: Change the comment as well as the type

src/Classes/ServiceAPI/MyRadio_Show.php
LloydW93: Yeah, that is right - it's an artifact of the data model being what MyRadioForm sends back.

PR #627, opened 2017-01-15
Updated nav bar

Brookke:
screenshot 2017-01-16 13 07 39

screenshot 2017-01-16 13 11 23

Updated the nav bar to bring it in with the 2016-site design


LordAro:
Pics? Some description of what you've done?

LordAro:
I'm reasonably happy with this, what are other people's thoughts on it?

I'd like to see the URY logo replaced with an SVG though, for consistency to match the MyRadio logo

Also, I wonder how obvious that the URY logo links back to the main website...

Brookke:
screenshot 2017-01-17 11 47 24
Just worth noting that when you hover over the URY logo it looks like this

Brookke:
Will replace with a SVG now

mstratford:
I think It may be sensible for a a "URY website" drop down style text on hover of the URY logo to make it more obvious.

Brookke:
Hmm I wonder how confusing it'll be in practice, because like all the other links on the nav bar when you hover over them they indicate that they are clickable (with the exception of the myradio logo, maybe that should change). With things like BBC iPlayer the BBC logo doesn't indicate that it goes back to the BBC home page, people just understand.

Brookke:
We could look at doing the text slide in thing to the right that you mentioned @mstratford providing it

mstratford:
In that case, it should be on the left like it will be on 201X-site :P with a divider?

Brookke:
That is another option @mstratford that I originally drew up and I am happy to move back to. My original thought to this design would mean that we can brand a URY site e.g. a blog neatly while clearly indicating is a seperate site and not part of the main URY site. Which having the logo on the left works too, but I thought this design was nicer. 

As to adding text on hover, I feel if things aren't clear when you first look at it then its not a good design and having text appear when hovering is just excessive and points out that the design is bad because its not intuitive to use.

src/Templates/parts/nav.twig
```diff
@@ -44,6 +39,11 @@
             </ul>
           </li>
         {% endif %}
+        {% if websiteurl %}
+        <li class="nav-img">
+          <a href="{{websiteurl}}"><img alt="MyRadio" src="{{baseurl}}img/{{shortname}}.svg"></a>
```
LordAro: alt text is wrong

(also, alt text/popup text is for accessibility as well)

Brookke: I will update the alt text. Yeah I get that alt text is for accessibility, but having a another element or otherwise being shown for non accessiblity reasons.

mstratford:
Just to remind you that the myradio button should hover background like the other links 👍

mstratford:
Also, should't it really be that the alt's for the images are labeled as "myradio Logo" and "URY Logo", with the links having title="current img alts here"?

mstratford:
Maybe Timeslot link should also have a title telling you it will "Select a timeslot" or something?

Brookke:
For the alt text I was following this https://www.w3.org/TR/html5/embedded-content-0.html#a-link-or-button-containing-nothing-but-an-image

mstratford:
Huh, fair enough.

Brookke:
We can add the titles in too though :)

Brookke:
@mstratford Shall we add the titles too?

mstratford:
Would be nice to have titles too :)

Brookke:
Will sort it out tomorrow :) Also drafted up this, if people would prefer it if it leads to less confusion
myradio

mstratford:
I'd prefer that to be honest :)

AJDurant:
I have to say, the word MyRadio looks pretty awful in roundabout

mstratford:
screen shot 2017-01-30 at 13 34 29
screen shot 2017-01-30 at 13 35 54

src/Templates/parts/nav.twig
```diff
@@ -7,16 +7,15 @@
         <span class="icon-bar"></span>
         <span class="icon-bar"></span>
       </button>
-      <a class="navbar-brand" href="{{baseurl}}">MyRadio</a>
+      {% if websiteurl %}
+        <a class="navbar-brand homepage-logo" href="{{websiteurl}}" title="Go to the {{shortname}} homepage"><img alt="Go to the {{shortname}} homepage" src="{{baseurl}}img/{{shortname}}.svg"></a>
+      {% endif %}
+      <span class="navbar-brand divider"></span>
+      <a class="navbar-brand myradio-logo" href="{{baseurl}}" title="Go to myradio main menu">
```
LordAro: Capitalisation of myradio here and the line below?

PR #628, opened 2017-01-16
Stop trying to sign in members more than once

LordAro:


src/Classes/ServiceAPI/MyRadio_Timeslot.php
```diff
@@ -1007,21 +1007,26 @@ public function sendMessage($message)
...
-        );
+        // If member already signed in for whatever reason, don't bother trying again.
+        $signedIn = !empty(self::$db->fetchOne(
+            'SELECT * FROM sis2.member_signin
+               WHERE show_season_timeslot_id=$1 AND memberid=$2',
+            [$this->getID(), $member->getID]
+        ));
+        if (!$signedIn) {
+            self::$db->query(
+                'INSERT INTO sis2.member_signin (show_season_timeslot_id, memberid, signerid)
```
AJDurant: Couldn't this just do INSERT OR IGNORE or something rather than 2 queries?

LordAro: Doesn't exist in postgres. There is UPSERT stuff in psql 9.5, but i don't think we actually depend on that yet...

PR #630, opened 2017-01-17
ESLint and other tidies

LordAro:
It's time I did a silly massive commit again, don't you think?

I wouldn't bother looking at the overall diff given the first commit which reformats essentially all of the JS files. The rest of the individual commits should closer represent what's going on

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -130,4 +130,12 @@
  <target name="phpunit" description="Run unit tests with PHPUnit">
   <exec executable="phpunit" failonerror="true"/>
  </target>
+
+ <target name="eslint" description="Lint JS with ESLint">
+  <exec executable="eslint">
+   <arg value="--config=${basedir}/.eslintrc.json" />
+   <arg value="--ignore-path=${basedir}/.eslintignore" />
+   <arg path="${basedir}/src/Public/js" />
+  </exec>
+ </target>
 </project>
--- a/build.xml
+++ b/build.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project name="MyRadio" default="build">
  <target name="build"
-   depends="prepare,lint,phploc,pdepend,phpmd-ci,phpcs-ci,phpcpd,phpdox,phpunit"/>
+   depends="prepare,lint,phploc,pdepend,phpmd-ci,phpcs-ci,phpcpd,phpdox,phpunit,eslint"/>
 
  <target name="build-parallel"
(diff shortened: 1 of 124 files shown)


src/Classes/Autoloader.php
```diff
@@ -3,7 +3,7 @@
 namespace MyRadio;
 
 /**
- * https://github.com/php-fig/fig-standards/blob/908fa291434a29951868b5111889ea6d0b37481a/accepted/PSR-4-autoloader-examples.md.
```
AJDurant: the reason for the blob id was so that the relevant version was linked

src/Public/js/myradio.core.js
```diff
@@ -1,141 +1,142 @@
...
-            url += '/' + firstParam;
-        }
-
-        $.ajax({
-            url: url,
-            data: options,
-            method: method
-        });
+/* global mConfig */
+var myradio = {
```
LordAro: I believe they are equivalent, as per http://stackoverflow.com/a/4862268/995325

src/Classes/Autoloader.php
LordAro: i get that, but the line was too long :p

also, it was "hiding" some fixes they've made over the last 3 years (very few, regardless)

build.xml
```diff
@@ -109,7 +109,7 @@
    <arg value="--report-checkstyle=${basedir}/build/logs/checkstyle.xml" />
    <arg value="--report-summary" />
    <arg value="--standard=PSR2" />
-   <arg value="--ignore=*.min.*,*/vendor/*,*/PublicAPI/rtfm/*" />
+   <arg value="--ignore=*.min.*,*/vendor/*,*/PublicAPI/rtfm/*,*/Public/js/*" />
    <arg path="${basedir}/src" />
    <arg value="-p" />
   </exec>
```
LordAro: Not sure if this should be added to the jenkins build, or the travis build. Thoughts?

AJDurant: Jenkins currently does all the linting, Travis does the tests? I guess we could keep it like that. I think it might be nice to consolidate into one platform, but separation may make sense. It would be nice if the Jenkins reports were more easily accessible though

LordAro: Done

LordAro:
Rebased

AJDurant:
You've added a target, but the target isn't called, and it needs to output an xml report in the 'checkstyle' format, which can then be read by jenkins.

AJDurant:
You should now in the project settings in Jenkins be able to add eslint thresholds under the validation plugin. And graphs.

Issue #85, opened 2014-06-18
Setup path for upgrades


LloydW93:
Currently, upgrading just puts you on to ??? after. This needs work.

LordAro:
https://gist.github.com/LordAro/ebb5a8c809dc355a2b528fbecde14e0b Diff between a vagrant install and production, as of this date

PR #632, opened 2017-01-19
Kill api_key_log table with fire

### LINKED ISSUE #344: Clean up log tables periodically
Migrated from phab
myury.error_rate and myury.api_key_log need to have some kind of system that cleans up old data periodically (3 months?).

LordAro:
Far too noisy, and taking up gigabytes of database storage

And no one ever looked at it anyway

Has the side effect of removing Config::exception_limit as well, which I'm fairly sure doesn't get hit anymore

Fixes #344


AJDurant:
I do think it would be nice to make the db auto update. It shouldn't actually be too hard to do, the main thing is to include SQL patch files to make the changes as well as changing the base SQL. We could start doing this before doing the automation anyway to help in deployment.

LordAro:
Oh I agree, I'd very much like to see myradio auto deploy on changes to master, of which this is the blocker (not including the existing differences between production db & base.sql)

However, out of scope for this prq :)

LordAro:
For reference, this is the diff of a fresh vagrant install vs production db:
https://gist.github.com/LordAro/ebb5a8c809dc355a2b528fbecde14e0b

LordAro:
(Rebased to leave exception_limit inplace, given it showed that it was still needed recently....)

LloydW93:
Would've been part of https://github.com/UniversityRadioYork/MyRadio/issues/85.

PR #636, opened 2017-01-22
Alter the width of items in forms

LordAro:
Think this improves things over forcing the width to 90% which makes dropdowns and so on rather wide


LordAro:
It doesn't look brilliant sure, but until the forms get properly bootstrapised, I'm not sure we can do much better. Feel free to fiddle around yourself though

This at least solves the wrapping issue with checkboxes

mstratford:
@LordAro Go for it then 👍

PR #638, opened 2017-01-24
Lintmore

LordAro:
Foobar.

One thing I'm not sure about - both joyride js files appear to have a mysterious constant if condition - https://github.com/UniversityRadioYork/MyRadio/blob/master/src/Public/js/myury.joyride.first_show.js#L8 & https://github.com/UniversityRadioYork/MyRadio/blob/master/src/Public/js/myury.joyride.banner.js#L8

They've had that condition as long as the files have been in the repo, but I can't work out why at all. Just a typo?

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -99,6 +99,7 @@
          description="Find coding standard violations using PHP_CodeSniffer and print human readable output. Intended for usage on the command line before committing.">
   <exec executable="phpcs">
    <arg value="--standard=PSR2" />
+   <arg value="--exclude=Squiz.Classes.ValidClassName.NotCamelCaps" />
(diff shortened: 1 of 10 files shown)


AJDurant:
Is {{action}} replaced somehow, or is it a template of an unfinished joyride? I don't really know anything about them @lloydw ?

LloydW93:
I suspect the joyride files used to be in the Twig templates directly and is now broken from being moved out. But I remember basically nothing about them.

LordAro:
I'll take them out since they apparently do nothing and haven't been needed anyway

(does joyride stuff even work anyway?)

LordAro:
I couldn't find anything anywhere, I'm happy to be proved wrong though

PR #639, opened 2017-01-25
Fix obit

LloydW93:
lock was renamed to setLock.


AJDurant:
@LloydW93 Can't write commit messages, and @adfw doesn't read or tidy up

PR #642, opened 2017-01-26
Add dev dependencies to composer config, and configure ant to use them

LordAro:


LordAro:
Be interesting to see how jenkins copes with that...

LordAro:
Should work now

PR #645, opened 2017-01-27
Edit History for Track Editor + Submit Blank Digitised By.

mstratford:
Will need db additions to rec_track.

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Track.php
+++ b/src/Classes/ServiceAPI/MyRadio_Track.php
@@ -532,13 +532,13 @@ class MyRadio_Track extends ServiceAPI
      */
     public function setLastEdited()
     {
-        $this->last_edited_time = time();
+        $this->last_edited_time = CoreUtils::getTimestamp();
         $this->last_edited_memberid = $_SESSION['memberid'];
         self::$db->query(
             'UPDATE rec_track SET last_edited_time=$1, last_edited_memberid=$2 WHERE trackid=$3',
             [
-                $this->last_edited_time ? CoreUtils::getTimestamp($this->last_edited_time) : null,
-                $_SESSION['memberid'],
+                $this->last_edited_time,
+                $this->last_edited_memberid,
                 $this->getID()
             ]);
         $this->updateCacheObject();
--- a/src/Classes/ServiceAPI/MyRadio_Track.php
+++ b/src/Classes/ServiceAPI/MyRadio_Track.php
@@ -273,7 +273,8 @@ class MyRadio_Track extends ServiceAPI
                 [
                     'label' => 'Digitised By',
                     'explanation' => 'The person who uploaded the track.',
-                    'enabled' => false
+                    'enabled' => false,
+                    'required' => false
                 ]
             )
         )->addField(
--- a/src/Classes/ServiceAPI/MyRadio_Track.php
+++ b/src/Classes/ServiceAPI/MyRadio_Track.php
@@ -111,6 +111,20 @@ class MyRadio_Track extends ServiceAPI
      */
     private $digitisedby;
 
+     /**
+     * The time when this track was last edited.
+     *
+     * @var int
+     */
+    private $lastedited;
+
+     /**
+     * The member who last edited this track.
+     *
+     * @var int
+     */
+    private $lasteditedby;
+
     /**
      * Caches Last.fm's Track.getSimilar response.
      *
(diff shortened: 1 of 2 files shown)


mstratford:
Doesn't appear to clear the cache for lasteditedby on the form, but does for lastedited, any ideas please :)

src/Classes/ServiceAPI/MyRadio_Track.php
```diff
@@ -412,6 +467,30 @@ public function getDigitisedBy()
...
+     * Get the user who last edited the track.
+     *
+     * @return MyRadio_User $lasteditedby The user who last edited the track.
+     */
+    public function getLastEditedBy()
+    {
+        if ($this->lasteditedby === null) {
+            return;
+        } else {
+            return MyRadio_User::getInstance($this->lastedited);
```
AJDurant: Isn't this the timestamp? I would recommend using clearer variable names, like last_edited_time and last_edited_memberid

src/Classes/ServiceAPI/MyRadio_Track.php
```diff
@@ -447,6 +526,24 @@ public function setDigitisedBy($digitisedby)
...
+     *
+     * @param bool $digitised
+     */
+    public function setLastEdited()
+    {
+        $this->lastedited = time();
+        self::$db->query(
+            'UPDATE rec_track SET lastedited=$1, lasteditedby=$2 WHERE trackid=$3',
+            [
+                $this->lastedited ? CoreUtils::getTimestamp($this->lastedited) : null,
```
AJDurant: unnecessary ternary given L535

src/Classes/ServiceAPI/MyRadio_Track.php
```diff
@@ -447,6 +526,24 @@ public function setDigitisedBy($digitisedby)
     }
 
     /**
+     * Update when a user last edited the track info.
+     *
+     * @param bool $digitised
+     */
+    public function setLastEdited()
+    {
+        $this->lastedited = time();
```
AJDurant: you need to update the editedby variable too.

mstratford: Ahh, that's what the problem was 👍 Cheers!

src/Classes/ServiceAPI/MyRadio_Track.php
mstratford: Looks like I can't read git diff...

src/Classes/ServiceAPI/MyRadio_Track.php
mstratford: Just tried removing the getTimestamp, database didn't like it.

src/Classes/ServiceAPI/MyRadio_Track.php
```diff
@@ -532,11 +532,12 @@ public function setDigitisedBy($digitisedby)
...
     {
-        $this->lastedited = time();
+        $this->last_edited_time = time();
+        $this->last_edited_memberid = $_SESSION['memberid'];
         self::$db->query(
-            'UPDATE rec_track SET lastedited=$1, lasteditedby=$2 WHERE trackid=$3',
+            'UPDATE rec_track SET last_edited_time=$1, last_edited_memberid=$2 WHERE trackid=$3',
             [
-                $this->lastedited ? CoreUtils::getTimestamp($this->lastedited) : null,
+                $this->last_edited_time ? CoreUtils::getTimestamp($this->last_edited_time) : null,
```
AJDurant: this ternary is still unnecessary, as you always set the variable. Not the conversion, but the `foo ? foo : bar`

src/Classes/ServiceAPI/MyRadio_Track.php
```diff
@@ -532,11 +532,12 @@ public function setDigitisedBy($digitisedby)
...
                 $_SESSION['memberid'],
```
AJDurant: it might be clearer to use $this->last_edited_memberid here

src/Classes/ServiceAPI/MyRadio_Track.php
```diff
@@ -163,8 +163,8 @@ protected function __construct($result)
         $this->clean = $result['clean'];
         $this->digitised = ($result['digitised'] == 't') ? true : false;
         $this->digitisedby = empty($result['digitisedby']) ? null : (int) $result['digitisedby'];
-        $this->lastedited = empty($result['lastedited']) ? null : $result['lastedited'];
-        $this->lasteditedby = empty($result['lasteditedby']) ? null : (int) $result['lasteditedby'];
+        $this->last_edited_time = empty($result['last_edited_time']) ? null : $result['last_edited_time'];
```
AJDurant: if you're converting it into a timestamp to go into the db, then do you need to convert here to store in the object?

PR #646, opened 2017-01-27
Updated base.sql with new last_edited storage.

mstratford:


schema/base.sql
```diff
@@ -6005,6 +6007,12 @@ ALTER TABLE ONLY strm_log
 ALTER TABLE ONLY rec_track
     ADD CONSTRAINT "$3" FOREIGN KEY (clean) REFERENCES rec_cleanlookup(clean_code) ON UPDATE RESTRICT ON DELETE RESTRICT;
 
+--
+-- Name: $4; Type: FK CONSTRAINT; Schema: public; Owner: myradio
+--
+
+ALTER TABLE ONLY rec_track
+    ADD CONSTRAINT "$4" FOREIGN KEY (last_edited_memberid) REFERENCES member(memberid) ON UPDATE CASCADE ON DELETE SET NULL;
```
AJDurant: The constraint names "$N" are from an older version of postgres, it would be better to give it a proper name
see: https://www.postgresql.org/message-id/19489.1126379421%40sss.pgh.pa.us

LordAro:
Added commit. Same name as in production

Issue #349, opened 2015-02-11
Show Planner: Show album name


AJDurant:
Show the album for tracks - probably in the tooltip

mstratford:
Will add :)

mstratford:
Implemented in #649

Issue #622, opened 2017-01-14
Library Search doesn't output clean/album correctly.


mstratford:
It seems that the data gets mangled going through MyRadio forms ->render(), since using the regular table renderer does display 'Clean' and 'Digitised' as True / False.

I've also noticed that for some reason, songs listed with 'Clean' shown as 1 are in fact marked as 'u' in the database (unknown cleanliness). Not sure if this is to do with the above issue, but it should be checked out also.

Lastly, albums come though as arrays from MyRadio_Track::findByOptions, so it doesn't display in the table.

PR #649, opened 2017-01-29
Fix up Show Planner fully (Including move to API)

mstratford:
Big changes to show planner's underside 👍


src/Public/js/nipsweb.librarypane.js
```diff
@@ -33,12 +33,16 @@ function updateCentralSearch()
         if (!data[file].clean) {
           classes = classes + " unclean";
         }
+        var cleanStars = "";
+        if (!data[file].clean) {
+          cleanStars = "**";
```
LloydW93: and a tooltip that explains what this means?

src/Public/js/nipsweb.player.js
```diff
@@ -94,7 +94,7 @@ var NIPSWeb = function (d) {
   };
 
   /**
-   * Change shipping operates in a queue - this ensures that changes are sent atomically and sequentially.
+   * Change shipping operates in a queue - this ensures that changes are sent automically and sequentially.
```
LloydW93: automically?

src/Templates/NIPSWeb/main.twig
```diff
@@ -7,7 +7,9 @@
 <div id="notice" class="alert alert-warning">Preparing show plan...</div>
 {# <div id="head-icons"><button class="btn-feedback">Send Feedback</button>&nbsp;<button class="btn-help"></button><button class="btn-logout" onclick="window.location = '/portal/redir_out.php'"></button></div> #}
 <header id="content-header">
-  <h3 id="a-manage-library"><a href="{{MyRadio.makeURL(config,'NIPSWeb','manage_library')}}" title="Library Manager">Manage Library</a></h3>
+  <div id="baps-menu" class="btn-group">
+    <a  href="{{MyRadio.makeURL(config,'NIPSWeb','manage_library')}}" id="menu-track-upload" class="btn btn-default"><span class="glyphicon glyphicon-upload"></span> Upload Tracks</a>
```
LloydW93: Not sure on the change of phrasing here - to me "Tracks" are things in the central library, not things like beds/jingles, which you can also manage here.

src/Public/js/nipsweb.librarypane.js
```diff
@@ -209,9 +216,9 @@ $(document).ready(function () {
   /**
    * Handler for activating the Manage Library link
    */
-  $("#a-manage-library").click(function () {
+  $("#menu-track-upload").click(function () {
     var url = $(this).children("a").attr("href");
-    myradio.createDialog("Manage Library", "<iframe src='" + url + "' width='580' height='500' frameborder='0'></iframe>");
+    myradio.createDialog("Upload Tracks", "<iframe src='" + url + "' width='580' height='500' frameborder='0'></iframe>");
```
AJDurant: It would be consistent to have 'Upload to Library' here too

src/Public/js/nipsweb.librarypane.js
```diff
@@ -30,15 +30,22 @@ function updateCentralSearch()
...
         }
         var classes = "";
+        var cleanStars = "";
+        var tooltip = "";
         if (!data[file].clean) {
           classes = classes + " unclean";
+          cleanStars = "**";
+          tooltip = "This track is explicit. Do not broadcast before 9PM."
+        }
+        if (!data[file].clean) {
```
LordAro: Empty if statement?

src/Public/js/nipsweb.librarypane.js
```diff
@@ -30,15 +30,22 @@ function updateCentralSearch()
...
+
         }
 
         $("#baps-channel-res").append(
           "<li id='" + data[file].album.recordid + "-" + data[file].trackid +
-          "' intro='" + data[file].intro + "'" +
+          "' intro='" + data[file].intro + "' title='" + tooltip + "'" +
```
LordAro: what happens here if `tooltip` isn't set? the title attribute is still set, does it still display something?

src/Public/js/nipsweb.librarypane.js
```diff
@@ -66,7 +67,7 @@ $(document).ready(function () {
       //Load a managed playlist
       $("#res-loading").show();
       $.ajax({
-        url: myradio.makeURL("MyRadio", "a-findtrack"),
+        url: mConfig.api_url + "/v2/track/search/",
```
AJDurant: This may be useful to you, or may need imporving, but it's probably worth using the API functionality from one place: https://github.com/UniversityRadioYork/MyRadio/blob/master/src/Public/js/myradio.core.js#L52

src/Public/js/nipsweb.librarypane.js
```diff
@@ -15,30 +15,34 @@ function updateCentralSearch()
...
-      require_digitised: true
-    },
+            url: mConfig.api_url + "/v2/track/search/",
+            type: 'get',
+            data: {
+              artist: $("#res-filter-artist").val(),
+              title: $("#res-filter-track").val(),
+              limit: 100,
+              digitised: true
+            },
```
LordAro: That's a lot of extra indentation (run eslint!)

AJDurant:
My [comment](https://github.com/UniversityRadioYork/MyRadio/pull/649#discussion_r98545842) pointing out there is a js function for calling the API already still stands: https://github.com/UniversityRadioYork/MyRadio/blob/master/src/Public/js/myradio.core.js#L52

src/Public/js/nipsweb.librarypane.js
```diff
@@ -1,3 +1,21 @@
...
+  '>': '&gt;',
+  '"': '&quot;',
+  "'": '&#39;',
+  '/': '&#x2F;',
+  '`': '&#x60;',
+  '=': '&#x3D;'
+};
+
+function escapeHTML (string) {
+  return String(string).replace(/[&<>"'`=\/]/g, function (s) {
```
mstratford: I just copied it from Stack Overflow :P

mstratford:
Fixes #349 and fixes #536

schema/patches/2.sql
```diff
@@ -0,0 +1 @@
+INSERT INTO api_class_map (class_name, api_name) VALUES ('\MyRadio\ServiceAPI\MyRadio_Artist', 'Artist');
```
LloydW93: Yeah, given it's also in base.sql, this shouldn't be here.

src/Classes/ServiceAPI/MyRadio_Track.php
```diff
@@ -872,6 +877,18 @@ public static function findByOptions($options)
                 ->getTracks()
             );
     }
+    /**
+     * This method allows direct access to iTonesPlaylists, without all the weird stuff in search api.$_COOKIE
+     *
+     * @param int itonesplaylistid The id of the itones playlist we want to get tracks for.
+    */
+    public static function itonesPlaylist($itonesplaylistid = null) {
```
LloydW93: If the intention is for an API endpoint, why not just expose iTones_Playlist->getTracks()?

src/Controllers/NIPSWeb/default.php
```diff
@@ -12,13 +12,14 @@
 
 CoreUtils::requireTimeslot();
 
+    $show_title = MyRadio_Timeslot::getInstance($_SESSION['timeslotid'])->getMeta('title');
+    $start_time = CoreUtils::HappyTime(MyRadio_Timeslot::getInstance($_SESSION['timeslotid'])->getStartTime());
```
LloydW93: happyTime

src/Classes/ServiceAPI/MyRadio_Track.php
```diff
@@ -676,9 +678,11 @@ public static function search(
             'digitised' => filter_var($digitised, FILTER_VALIDATE_BOOLEAN),
             'clean' => $clean,
             'precise' => filter_var($precise, FILTER_VALIDATE_BOOLEAN),
-            'limit' => (($page - 1) * 50) . ',50'
+            'itonesplaylistid' => $itonesplaylistid
         ];
-
+        if ($limit != null) {
```
LloydW93: Why's pagination being replace with a limit?

src/Controllers/NIPSWeb/default.php
```diff
@@ -30,7 +31,9 @@
 }
 
 CoreUtils::getTemplateObject()->setTemplate($template)
-    ->addVariable('title', $title)
+    ->addVariable('title', "Show Planner")
+    ->addVariable('show_title', $show_title)
```
LloydW93: timeslotname is currently globally exposed - is it worth doing so with these and removing timeslotname, instead of having kind-of-duplication?

src/Controllers/NIPSWeb/import.php
```diff
@@ -0,0 +1,11 @@
+<?php
+/**
+ * Main renderer for NIPSWeb.
```
LloydW93: Is not :p

src/Public/js/myradio.core.js
```diff
@@ -49,7 +49,15 @@ var myradio = {
     );
     return reportButton;
   },
-  callAPI: function (method, module, action, id, firstParam, options) {
+  callAPI: function (method, module, action, id, firstParam, options, successFunc = function(){}) {
```
LloydW93: Default parameter support is not supported in Safari < 10, or any version Opera. (or IE, but meh). Discuss whether this is acceptable minimum browser support and change if not.

src/Public/js/nipsweb.import.js
```diff
@@ -0,0 +1,265 @@
...
+  return results[1] || 0;
+};
+var nextWeightChannel0;
+var nextWeightChannel1;
+var nextWeightChannel2;
+
+$(document).ready(
+  function () {
+
+    nextWeightChannel0 = parseInt($.urlParam("channel0lastweight"))+1;
```
LloydW93: parseInt should always be used with a radix parameter - the spec does not actually define a default number base!

src/Public/js/nipsweb.import.js
```diff
@@ -0,0 +1,265 @@
...
+    nextWeightChannel1 = parseInt($.urlParam("channel1lastweight"))+1;
+    nextWeightChannel2 = parseInt($.urlParam("channel2lastweight"))+1;
+
+    $.ajax({
```
LloydW93: Can't this use callAPI?

src/Public/js/nipsweb.import.js
```diff
@@ -0,0 +1,265 @@
...
+
+function selectChannel(channelNo) {
+  $("#import-channel-selector a").removeClass("active");
+  $("#import-channel-" + channelNo).addClass("active");
+  loadChannelList();
+}
+
+function loadChannelList() {
+  var selectedTimeslotID = $("#import-timeslot-selector").find(":selected").attr("value");
+  $.ajax({
```
LloydW93: As with other, why can't this use callAPI?

src/Public/js/nipsweb.librarypane.js
```diff
@@ -1,52 +1,75 @@
-/* global Bloodhound, myradio, planner, ulsort */
+/* global Bloodhound, myradio, planner, ulsort, showAlert */
 /**
  * This file contains the necessary functions for browsing searching and adding
  * items in the Library pane of the main interface
  * @todo Includes length attribute to list items
  */
 
+//Still can't believe js/jquery doesn't have escaping built-in!
```
LloydW93: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent?

src/Public/js/nipsweb.librarypane.js
```diff
@@ -143,36 +168,54 @@ $(document).ready(function () {
...
-      datumTokenizer: Bloodhound.tokenizers.obj.whitespace("title"),
+      datumTokenizer: function (datum) {
+        return Bloodhound.tokenizers.whitespace(datum.title);
+      },
       queryTokenizer: Bloodhound.tokenizers.whitespace,
       limit: 5,
       dupDetector: function (remote, local) {
         return local.title == remote.title;
       },
       prefetch: {
```
LloydW93: We've previously discussed that prefetch could be the cause of some of our issues with search. Maybe pull that block out?

src/Public/js/nipsweb.player.js
```diff
@@ -1,4 +1,35 @@
...
+  }).addClass("alert-"+type).html(icon + text);
+};
+
+//reload the page after something good or bad happens
+var reload = function() {
+  location.reload();
+};
+
+// Stores the clientid to enable multiple editors
+var clientid = null;
```
LloydW93: What's with hoisting this up to global scope?

src/Public/js/nipsweb.player.js
```diff
@@ -1,4 +1,35 @@
-/* global ChannelConfigurator, myradio, mConfig */
+/* global ChannelConfigurator, myradio */
+//Global popup alert controller for all nipsweb js files.
+var showAlert = function (text, type) {
```
LloydW93: This too (why global?)

src/Templates/NIPSWeb/import.twig
```diff
@@ -0,0 +1,42 @@
+<!DOCTYPE html>
```
LloydW93: I'm wondering how easy it'd be to deduplicate the stuff that's been copy/pasted from the general timeslot selector. They both just ultimately spit out a timeslotid, but do something different with it, right?

src/Templates/parts/base_foot.twig
```diff
@@ -3,5 +3,7 @@
 <script type="text/javascript">
   window.myradio.module = "{{module}}";
   window.myradio.action = "{{action}}";
+  window.memberid = parseInt({{memberid}});
```
LloydW93: Maybe keep these inside the window.myradio object (limit things directly in global scope).

LloydW93:
A clean install is missing access to https://localhost:4443/api/v2/user/1/shows/.

LloydW93:
The context menu now only works on items, not blank areas of a channel.

src/Classes/ServiceAPI/MyRadio_Track.php
mstratford: Erm, mainly because it's current only use was for show planner, which likes limits of results 👍

src/Classes/ServiceAPI/MyRadio_Track.php
mstratford: Hmm, Wasn't really in the mood for all the stuff regarding adding another API endpoint that way :( This was supposed to be quick, and I would finish API move during summer :(

src/Controllers/NIPSWeb/import.php
mstratford: oops :p

src/Classes/ServiceAPI/MyRadio_Track.php
LloydW93: But surely you're just adding a new API endpoint either way, so it was the same amount of work? :p

src/Classes/ServiceAPI/MyRadio_Track.php
LloydW93: it's 'better' to get do something like get the first 50 and expand (like infinite scrolling, but I hate that too). But meh.

src/Templates/NIPSWeb/import.twig
mstratford: Please don't make me do this 👎  😄

src/Public/js/myradio.core.js
```diff
@@ -49,7 +49,10 @@ var myradio = {
     );
     return reportButton;
   },
-  callAPI: function (method, module, action, id, firstParam, options, successFunc = function(){}) {
+  callAPI: function (method, module, action, id, firstParam, options, successFunc) {
+    if (successFunc == undefined) {
```
LloydW93: "typeof successFunc === 'undefined'" would be better - technically, this will throw an error, e.g.:
```
chicken == undefined
VM1569:1 Uncaught ReferenceError: chicken is not defined
    at <anonymous>:1:1
```

mstratford: Ok :)

src/Classes/ServiceAPI/MyRadio_Track.php
mstratford: I'll come back to it when I make the library search page actually use the API :)

src/Public/js/nipsweb.import.js
mstratford: Well, there's probably some lingering examples of this then elsewhere, since I copied it from somewhere, although I just got rid of some :)

src/Public/js/nipsweb.player.js
mstratford: Oh, it was used by other stuff, before I made those things use the API 👍

src/Classes/ServiceAPI/MyRadio_Track.php
mstratford: #622 to remind me 👍

src/Controllers/NIPSWeb/default.php
mstratford: Not sure quite what you mean here. Sorry.

src/Public/js/nipsweb.librarypane.js
mstratford: Maybe I'm using the wrong term, I mean converting into html special characters like &amp; from &, since the tooltips won't work with ' in the titles (closes attribute and keeps going).

There tended to be quite a bit of confusion among the interwebs with this one, and I've been directed to that link before. Using that just gives me %23 etc within the displayed tooltip.

src/Templates/NIPSWeb/import.twig
LloydW93: I raised #676 instead.

PR #651, opened 2017-01-29
Resync schema.sql with what's currently live

LordAro:
It's not perfect, but is much better, I think.

Brief incomplete summary:
* Little bit of function reformatting
* Removed a bit of public rapier schema
* Removed site_package/sitetree schemas - appears to have been a precursor to metadata?
* Added some missing constraints
* Deleted/Moved sis_* tables to sis2 schema and renamed appropriately. Have also done this on live db

### DIFF
--- a/schema/base.sql
+++ b/schema/base.sql
@@ -15,7 +15,7 @@ COMMENT ON SCHEMA people IS 'Tables for the LeRouge Extensible Roles (Or User Gr
 CREATE SCHEMA schedule;
 COMMENT ON SCHEMA schedule IS 'Schema for the MyRadio schedule.';
 CREATE SCHEMA sis2;
-COMMENT ON SCHEMA sis2 IS 'Used by Studio Infomation Service. Deprecated somewhat.';
+COMMENT ON SCHEMA sis2 IS 'Used by Studio Infomation Service.';
 CREATE SCHEMA tracklist;
 COMMENT ON SCHEMA tracklist IS 'Provides a schema that logs played out tracks for PPL track returns';
 CREATE SCHEMA uryplayer;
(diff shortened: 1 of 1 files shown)


LordAro:
( https://gist.github.com/LordAro/af59c6fd1c6ed9c17414272a9e082799 used as reference - schema.sql is taken from live, schema2.sql taken from a fresh myradio install, both generated with `pg_dump -s -O -x` )

schema/base.sql
```diff
@@ -243,6 +248,15 @@ CREATE SEQUENCE bapsplanner.timeslot_items_timeslot_item_id_seq
     CACHE 1;
 ALTER SEQUENCE bapsplanner.timeslot_items_timeslot_item_id_seq OWNED BY bapsplanner.timeslot_items.timeslot_item_id;
 SET search_path = jukebox, pg_catalog;
+CREATE TABLE playlist_availability (
+    memberid integer,
```
LloydW93: Indent

schema/base.sql
```diff
@@ -243,6 +248,15 @@ CREATE SEQUENCE bapsplanner.timeslot_items_timeslot_item_id_seq
...
+	approvedid integer,
+	effective_from timestamp with time zone NOT NULL,
+	effective_to timestamp with time zone,
+	playlist_availability_id integer DEFAULT nextval('website.banner_campaign_banner_campaign_id_seq'::regclass) NOT NULL,
```
LloydW93: Dat sequence...

schema/base.sql
```diff
@@ -6091,15 +6056,15 @@ ALTER TABLE ONLY auth_user_groups
 --
 
 ALTER TABLE ONLY baps_item
-    ADD CONSTRAINT baps_item_fileitemid_fkey FOREIGN KEY (fileitemid) REFERENCES baps_fileitem(fileitemid);
+    ADD CONSTRAINT baps_item_fileitemid_fkey FOREIGN KEY (fileitemid) REFERENCES baps_fileitem(fileitemi) ON UPDATE CASCADE ON DELETE CASCADE;
```
LloydW93: fileitemi?

schema/base.sql
```diff
@@ -6309,6 +6274,13 @@ ALTER TABLE ONLY rec_itunes
 ALTER TABLE ONLY rec_track
     ADD CONSTRAINT rec_track_digitisedby_fkey FOREIGN KEY (digitisedby) REFERENCES member(memberid) ON UPDATE CASCADE ON DELETE SET NULL;
 
+--
+-- Name: rec_track_lasteditedby_fkey; Type: FK CONSTRAINT; Schema: public; Owner: -
+--
+
+ALTER TABLE ONLY rec_track
+    ADD CONSTRAINT rec_track_lasteditedby_fkey FOREIGN KEY (last_edited_memberid) REFERENCES member(memberid) ON UPDATE CASCADE ON DELETE SET NULL;
```
LloydW93: The db schema generally prefers to fail than silently change other data.

LordAro: This is what I found in the live schema ¯\\\_(ツ)\_/¯

schema/base.sql
LordAro: What of it? again, this is what the live schema spat out...

schema/base.sql
AJDurant: this does make sense. ON DELETE SET NULL ensures that if a member disappears then the track won't. This should also be on editedby

LloydW93: I'm pretty sure we block, i.e. the DELETE query errors, rather than CASCADE elsewhere. Imo, if a member has ever done something, then that trail shouldn't be deleted. An account can be deactivated if necessary... assuming that still works...

AJDurant: I agree with keeping the trail and not deleting members, but I do think more of the db should work without errors.

LordAro: So.. any consensus?

PR #654, opened 2017-01-31
Support for PHP 7

### LINKED ISSUE #526: Support PHP 7
http://php.net/manual/en/migration70.incompatible.php
- [x] Validate use of indirect variables/properties/methods (e.g. $$thing)
- [x] Foreaches by-reference where necessary
- [x] Use of mcrypt aliases
- [x] Fixes to Custom Session Handler Return Values

LloydW93:
Fixes #526. Turned out that PHP 7 itself was a case of Just Works, but Twig 2.0 was a real pain.


src/Controllers/Setup/checks.php
```diff
@@ -247,16 +240,10 @@ function convertPHPSizeToBytes($sSize)
...
             <code>
-              apt-get install php5-curl php5-geoip php5-gd php5-ldap php5-mcrypt php5-pgsql php5-dev php-pear<br>
-              pear channel-discover pear.twig-project.org<br>
-              pear install twig/Twig<br>
-              pear install twig/CTwig<br>
-              echo "extension=mcrypt.so" > /etc/php5/mods-available/mcrypt.ini<br>
-              ln -s /etc/php5/mods-available/mcrypt.ini /etc/php5/apache2/conf.d/20-mcrypt.ini<br>
-              echo "extension=twig.so" > /etc/php5/mods-available/twig.ini<br>
-              ln -s /etc/php5/mods-available/twig.ini /etc/php5/apache2/conf.d/20-twig.ini<br>
+              apt install php-curl php-geoip php-gd php-ldap php-pgsql php-dev composer<br>
```
LloydW93: Need to add php-mbstring

LloydW93:
Nope :(

.travis.yml
```diff
@@ -1,8 +1,14 @@
 language: php
 php:
     - '5.6'
```
LloydW93: We need to drop 5.6 as Twig 2 only supports 7.

LordAro: yeah... guess there's absolutely no way to support both? be a shame to completely drop 5.6

AJDurant:
Still need to make the `apt install` line in checks.php match up

PR #664, opened 2017-03-15
Deduplicate encode track logic

LordAro:


src/Classes/NIPSWeb/NIPSWeb_ManagedItem.php
```diff
@@ -227,31 +230,13 @@ public static function storeItem($tmpid, $title)
          * 3- Original file for potential future conversions
          */
         $tmpfile = Config::$audio_upload_tmp_dir.'/'.$tmpid;
-
-        if (!$item->getFolder()) {
-            //Creating folders failed.
-            return ['status' => 'FAIL', 'error' => 'Folders could not be created.', 'fileid' => $_REQUEST['fileid']];
-        }
         $dbfile = $item->getFolder().'/'.$item->getID();
+        CoreUtils::encodeTrack($tmpfile, $dbfile);
```
DinCahill: Why two calls to `encodeTrack`?

LordAro: because i can't read diffs properly

src/Classes/MyRadio/CoreUtils.php
```diff
@@ -233,6 +233,42 @@ public static function makeInterval($start, $end)
...
+    {
+        $commands = [
+            'mp3' => "nice -n 15 ffmpeg -i '{$tmpfile}' -ab 192k -f mp3 -map 0:a '{$dbfile}.mp3'",
+            'ogg' => "nice -n 15 ffmpeg -i '{$tmpfile}' -acodec libvorbis -ab 192k -map 0:a '{$dbfile}.ogg'"
+        ];
+        $escaped_commands = array_map('escapeshellcmd', $commands);
+        $failed_formats = [];
+
+        foreach ($escaped_commands as $format => $command) {
+            exec($command, $command_stdout, $command_exit_code);
```
LordAro: where do the stdout/stderr variables come from? where are they used?

mstratford:
#673 Maybe something to fix here? Not sure if we want to open the question to the Track Uploader UI or not? Would be useful to be able to simply replace tracks, but not sure if this should go under the edit form instead?

LordAro:
We could either change the ffmpeg invocation so that it'll overwrite files (`-y`) or return an error if the file already exists

Not sure which is best. Should certainly never happen for database entries, given the name of the file is the primary key, but manageditems might be a bit different...

src/Classes/MyRadio/CoreUtils.php
LloydW93: http://php.net/manual/en/function.exec.php - these are passed by reference and therefore defined by exec.

LloydW93:
I'm a bit concerned about CoreUtils just becoming a 'stuff' class.

Well, it already is, but we should probably try not to continue this.

LordAro:
It's less of a "stuff" class than it used to be! URLUtils and AuthUtils are things these days. I'm not sure where else it would go, is all.

Alternatively, splitting CoreUtils further? ideas on what should be split out?

PR #670, opened 2017-03-30
Fixed broken error redirect

mstratford:
Stops the 404, and also alerts us to places where this shouldn't be happening 👍

### DIFF
--- a/src/Classes/MyRadio/URLUtils.php
+++ b/src/Classes/MyRadio/URLUtils.php
@@ -82,6 +82,16 @@ class URLUtils
     }
 
     /**
+     * Redirects to another page, specified already by the caller.
+     *
+     * @param string  $URI The relative URI to redirect to.
+     */
+    public static function redirectURI($URI)
+    {
+        header('Location: '.$URI);
+    }
+
+    /**
      * Builds a module/action URL.
      *
      * @param string $module
--- a/src/Controllers/MyRadio/login.php
+++ b/src/Controllers/MyRadio/login.php
@@ -17,7 +17,6 @@ use \MyRadio\MyRadio\MyRadioFormField;
 
 if (isset($_SESSION['memberid'])) {
     isset($_GET["next"]) ? URLUtils::redirectURI($_GET["next"]) : URLUtils::redirect(Config::$default_module);
-
 } else {
     $form = (
         new MyRadioForm(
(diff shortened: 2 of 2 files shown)


src/Controllers/MyRadio/login.php
```diff
@@ -16,7 +16,7 @@
  */
 
 if (isset($_SESSION['memberid'])) {
-    URLUtils::redirect(Config::$base_url);
+    URLUtils::redirectWithMessage("/", "", "It seems you were trying to login while still being logged in, if that is not what happened, please contact the Computing Team.");
```
AJDurant: `"/", "",`
Doesn't really look right... Does this actually go to / or to /myradio/ or what is configured?

mstratford:
It goes to the menu, since redirect() is a module/action relative function.

src/Controllers/MyRadio/login.php
LordAro: Not sure I quite agree with the message. Why bother telling people?

mstratford: Meh, it was mainly because otherwise you've just redirected to the homepage for an unknown reason. It hits this when the API is being weird (like now).

mstratford: It goes to the menu (/myradio or /myury-dev etc), since redirect() is a module/action relative function.

mstratford: The JS (somewhat) rightfully assumes that if the API returns that it hasn't got valid credentials from a API call with JS myradio, then there must not be someone logged in. However, if the API is weird (like now), it will try to goto login whilst still actually logged in, so lands at the homepage.

mstratford:
Right, fixed :)

src/Classes/MyRadio/URLUtils.php
```diff
@@ -82,6 +82,16 @@ public static function redirectWithMessage($module, $action, $message)
     }
 
     /**
+     * Redirects to another page, specified already by the caller.
+     *
+     * @param string  $URI The relative URI to redirect to.
+     */
+    public static function redirectURI($URI)
+    {
+        header('Location: '.$URI);
```
LloydW93: Is it worth restricting what this can redirect to (e.g. to ury.org.uk) in case of something evil? Probably not...

mstratford: Hmm, I think if someone wanted to go to that page anyway, then there's easier ways of doing it :p

Issue #668, opened 2017-03-26
Support PHP 7.1


LloydW93:
#526 is done, but what actually got deployed this week is 7.1 (yes, I edited the issue title of the original).

7.1 introduces backwards incompatible changes which we should review (an initial look seems fine), "other changes" which are in fact also backwards incompatible (Show Planner broke), and a set of new warnings for upcoming deprecations and general strictness that need to go away.

LordAro:
http://php.net/manual/en/migration71.incompatible.php link for helpfulness

LordAro:
I believe we've got all these now. Anyone noticed anything else?
@mstratford @LloydW93

mstratford:
I don't think I've seen any warnings in a while.

mstratford:
There's one while editing banners, but we're working on this already.

mstratford:
Also one in Charts.

LordAro:
```
Warning : Declaration of MyRadio\ServiceAPI\MyRadio_Banner::create(MyRadio\ServiceAPI\MyRadio_Photo $photo, $alt = 'Unnamed Ba...', $target = NULL, $type = 2) should be compatible with MyRadio\ServiceAPI\MyRadio_Photo::create($tmp_file) - In /usr/local/www/myradio/src/Classes/ServiceAPI/MyRadio_Banner.php on line 19
```

```
Unknown error code : Non-static method MyRadio\ServiceAPI\MyRadio_ChartType::getAll() should not be called statically - In /usr/local/www/myradio/src/Controllers/Charts/default.php on line 9
```

For reference porpoises. Fairly sure the second on is not new

LordAro:
Fixed

PR #672, opened 2017-03-31
Change the session serialization function to be the built-in serialize…

LloydW93:
…/unserialize methods, instead of slightly different internal ones, which enables the use of unserialize() to get at a session, instead of doing evil things to the  superglobal.


LloydW93:
Yep, this leaves DISABLE_SESSION working. The thing before was a hack, so I'm guessing the change in PHP7/7.1 was about not letting me do stupid things to $_SESSION anymore, which seeing as it's no longer necessary, is fair enough.

LordAro:
Comment to link it to #668

PR #680, opened 2017-04-01
Fix up some of the Stats pages.

mstratford:
Added / fixed some incorrect / missing datatables config js files and allowed Manual SIS entered tracks to not break the CSV/fullTracklist output.

Also removed some warnings from the top of the page!

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_TracklistItem.php
+++ b/src/Classes/ServiceAPI/MyRadio_TracklistItem.php
@@ -187,13 +187,11 @@ class MyRadio_TracklistItem extends ServiceAPI
             unset($data['digitised']);
             unset($data['deletelink']);
             unset($data['trackno']);
+            unset($data['intro']);
 
             //for manual SIS entries
             if (!isset($data['trackid'])) {
-                $data['trackid'] = 0;
-            }
-            if (!isset($data['intro'])) {
-                $data['intro'] = 0;
+                $data['trackid'] = "SIS Manual";
             }
 
             if (is_array($data['album'])) {
--- a/src/Classes/ServiceAPI/MyRadio_TracklistItem.php
+++ b/src/Classes/ServiceAPI/MyRadio_TracklistItem.php
@@ -188,6 +188,14 @@ class MyRadio_TracklistItem extends ServiceAPI
             unset($data['deletelink']);
             unset($data['trackno']);
 
+            //for manual SIS entries
+            if (!isset($data['trackid'])) {
+                $data['trackid'] = 0;
+            }
+            if (!isset($data['intro'])) {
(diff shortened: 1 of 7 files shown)


src/Controllers/Stats/fullTracklist.php
```diff
@@ -46,9 +46,9 @@
                 'Stats',
                 'fullTracklist',
                 [
-                    'rangesel-starttime' => array_key_exists($_GET['rangesel-starttime']) ?
+                    'rangesel-starttime' => isset($_GET['rangesel-starttime']) ?
                         $_GET['rangesel-starttime'] : '',
-                    'rangesel-endtime' => array_key_exists($_GET['rangesel-endtime']) ? $_GET['rangesel-endtime'] : '',
+                    'rangesel-endtime' => isset($_GET['rangesel-endtime']) ? $_GET['rangesel-endtime'] : '',
```
LordAro: there's quite a lot of this type of thing scattered around the codebase, but perhaps you could use some null coalescing operators here & above? http://stackoverflow.com/a/18603279/995325

src/Classes/ServiceAPI/MyRadio_TracklistItem.php
```diff
@@ -187,6 +187,12 @@ public static function getTracklistForTime($start, $end = null, $include_playout
             unset($data['digitised']);
             unset($data['deletelink']);
             unset($data['trackno']);
+            unset($data['intro']);
+
+            //for manual SIS entries
+            if (!isset($data['trackid'])) {
+                $data['trackid'] = "SIS Manual";
```
LordAro: Is the fact that this isn't an integer going to have implications? perhaps -1 or 0 would be better?

src/Controllers/Stats/fullTracklist.php
mstratford: Hmm, I guess, but I wanted to make it obvious. It's only used in the fullTrackList / csv from what I can see.

src/Classes/ServiceAPI/MyRadio_TracklistItem.php
mstratford: Hmm, I guess, but I wanted to make it obvious. It's only used in the fullTrackList / csv from what I can see.

src/Controllers/Stats/fullTracklist.php
mstratford: Ok, cool :)

src/Classes/ServiceAPI/MyRadio_TracklistItem.php
LloydW93: This is basically the general MyRadio problem of the layers being all mushed together.

PR #682, opened 2017-04-02
More minor fixinations

LordAro:
Again, looking at individual commits probably best


src/Public/js/nipsweb.librarypane.js
```diff
@@ -10,7 +10,7 @@ var entityMap = {
   "&": "&amp;",
   "<": "&lt;",
   ">": "&gt;",
-  '"': "&quot;",
+  "\"": "&quot;",
```
mstratford: Huh, I tried this before and my syntax highlighting went weird, but if it works, it works :)

LordAro: Then blame the syntax highlighter :p
ideally i would use '' for it, but eslint complained, so meh

src/Classes/ServiceAPI/MyRadio_BannerCampaign.php
```diff
@@ -494,44 +239,26 @@ public static function getCampaignLocations()
...
      * Returns the form needed to create or edit Banner Campaigns.
      *
-     * @param int $bannerid The ID of the Banner that this Campaign will be/is linked to
+     * @param int $newBannerID The ID of the Banner that this Campaign will be created for
+     * @param int $availabilityid The ID of the campaign that will be edited
      *
      * @return MyRadioForm
      */
-    public static function getForm($bannerid = null)
+    public static function getForm($availabilityid = null, $newBannerID = null)
```
LloydW93: I think use of camelCase for variable names is inconsistent with everything else (although tbf everything else is a mixture of _ and no _)

mstratford:
Well, I tested it, and found it still borked.

src/Public/js/nipsweb.import.js
```diff
@@ -1,4 +1,5 @@
 /* global myradio */
+/* exported importSelectedTracks,selectNone,SelectAll,reload */
```
mstratford: selectAll, not SelectAll (also fixed in my own patches)

src/Classes/MyRadio/MyRadio_Availability.php
```diff
@@ -310,24 +311,32 @@ public function setEffectiveTo($time)
...
      * @todo  Input validation.
      */
-    public function addTimeslot($day, $start, $end)
+    public function addTimeslot($day, $start, $end, $optional_field_name, $optional_field_val = NULL)
     {
         $start = gmdate('H:i:s', $start).'+00';
         $end = gmdate('H:i:s', $end).'+00';
+        $userid = MyRadio_User::getInstance()->getID();
+        if (!is_null($optional_field_name) && !is_null($optional_field_val)) {
+            $optional_field = true;
```
LordAro: boolean logic here

src/Classes/ServiceAPI/MyRadio_BannerCampaign.php
```diff
@@ -168,9 +168,9 @@ public static function create(
         $campaign = self::getInstance($result[0]);
 
         foreach ($timeslots as $timeslot) {
-            $campaign->addTimeslot($timeslot['day'], $timeslot['start_time'], $timeslot['end_time']);
+            $campaign->addTimeslot($timeslot['day'], $timeslot['start_time'], $timeslot['end_time'], 'order',$result[0]);
```
LordAro: is there an issue with '"order"' vs 'order' here? (i'd prefer to keep the code the same and change the column name instead)

src/Classes/MyRadio/MyRadio_Availability.php
```diff
@@ -310,24 +311,32 @@ public function setEffectiveTo($time)
      *
      * @todo  Input validation.
      */
-    public function addTimeslot($day, $start, $end)
+    public function addTimeslot($day, $start, $end, $optional_field_name, $optional_field_val = NULL)
```
LordAro: document what these optional fields do pls

src/Classes/ServiceAPI/MyRadio_BannerCampaign.php
mstratford: The column is just order in the db, without any quotes...

Issue #612, opened 2017-01-08
Setting an Intro doesn't bust the cache for a TimeslotItem


LloydW93:
NIPSWeb_TimeslotItem currently stores the $item object itself within it. This means that for the lifetime of the TimeslitItem, changes to the Track are not propagated. This should probably store a type/id pair here, and have the actual object for using in a property that doesn't get cached.

mstratford:
This is still a problem I think, setting into in Showplanner updates the edit track form, but not the show planner after reload?

LloydW93:
Never said it was fixed :)

PR #689, opened 2017-04-13
A few UI tweaks for New Show Planner

mstratford:
Also fixes the Multi Output audio selection in newer Chrome versions.


LordAro:
Will do a full review later, but did you fix the "attempting to reorder search results causes error" issue? Don't see it explicitly mentioned in commits, is all

mstratford:
@LordAro It was the first commit 👍

mstratford:
Fixes #612

mstratford:
Note, the length field in bapsplanner.managed_items will need to have the decimal points removed before this is updated.

LordAro:
Why can you not format the time accordingly? the microsecond accuracy is built into postgres' column type and isn't all that easy to (permanently) remove

schema/data-actions.json
```diff
@@ -71,7 +71,6 @@
     ["NIPSWeb", "secure_play"],
     ["NIPSWeb", "managed_play"],
     ["NIPSWeb", "load_aux_lib"],
-    ["NIPSWeb", "get_client_token"],
```
LordAro: Think this means something needs to be removed from the database?

src/Classes/NIPSWeb/NIPSWeb_ManagedItem.php
```diff
@@ -151,7 +148,7 @@ public function toDataSource($mixins = [])
             'summary' => $this->getTitle(), //Again, freaking NIPSWeb
             'title' => $this->getTitle(),
             'managedid' => $this->getID(),
-            'length' => CoreUtils::happyTime($this->getLength() > 0 ? $this->getLength() : 0, true, false),
+            'length' => $this->getLength(),
```
LordAro: an int() here would do the job on truncating the decimal points, right?

schema/data-actions.json
mstratford: I guess so, I can do that :)

src/Classes/NIPSWeb/NIPSWeb_ManagedItem.php
mstratford: Well, it comes out as a hh:mm:ss.ms so I'm not sure int will work. However I will add a substring thing here.

mstratford:
This doesn't fix / work around the DELETE issue, I don't know what's causing that in the URY environment.

mstratford:
Removed the need for the SQL decimal point edit using the above commit.

src/Classes/NIPSWeb/NIPSWeb_ManagedItem.php
```diff
@@ -72,7 +72,7 @@ protected function __construct($resid, $playlistref = null)
                 );
         $this->folder = $result['folder'];
         $this->title = $result['title'];
-        $this->length = $result['length'];
+        $this->length = strtotime('1970-01-01 '.$result['length']. ' UTC');
```
LordAro: i feel like there should be a timeToInt function somewhere...

schema/data-actions.json
mstratford: Done

PR #691, opened 2017-04-17
2 Random fixes

mstratford:


mstratford:
It didn't seem to be picking up the trackid field previously, so copied bits from elsewhere.

mstratford:
Might be to do with how "Number" is the first col of rec_track, not trackid.

src/Classes/ServiceAPI/MyRadio_Track.php
```diff
@@ -1289,11 +1289,11 @@ public function setIntro($duration)
     public static function getAllDigitised()
     {
         self::initDB();
-        $result = self::$db->fetchColumn(self::BASE_TRACK_SQL.' WHERE digitised=\'t\'');
```
LordAro: BASE_TRACK_SQL appears to only be used on one other place - get rid of it?

src/Controllers/Mail/send.php
```diff
@@ -27,14 +27,17 @@
...
         ]
     )
 )->addField(
     new MyRadioFormField(
         'body',
         MyRadioFormField::TYPE_BLOCKTEXT,
-        []
+        [
+            'label' => 'Message Body',
+        ]
```
LordAro: doesn't really matter, but squash these 3 onto 1 line?

PR #693, opened 2017-04-23
Find tracks better in bloodhound menus

mstratford:
To fix problems with play listing where less useful (most likely ordered by oldest (lowest trackid) first) results caused the required track to be not selectable.


mstratford:
From what i've seen so far, nothing with the MyRadio Form auto fill require undigitised tracks. If there is, and there's a better way to do the require digitised, then let me know.

mstratford:
Fixes #446

PR #695, opened 2017-04-24
Other stuff from #682

LordAro:


AJDurant:
Not reloading the page was deliberate, that is why there was preventDefault. These calls were for AJAX requests, the page display should have reacted itself.

LordAro:
Hrm, OK. How is the page display supposed to react?

AJDurant:
It entirely depends on what the action is. If you are revoking permissions for example then the permission element in the list should be removed from the page.

LordAro:
OK, removed for now, will work it out later

PR #702, opened 2017-05-29
Variety Up & Meeting PPL

Eiron:
This is a slight modification to increase variety on jukebox, whilst ensuring that the updated terms of our PPL license are adhered to.

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_TracklistItem.php
+++ b/src/Classes/ServiceAPI/MyRadio_TracklistItem.php
@@ -344,9 +344,10 @@ class MyRadio_TracklistItem extends ServiceAPI
      * Check whether queuing the given Track for playout right now would be a
      * breach of our PPL Licence.
      *
-     * The PPL Licence states that a maximum of two songs from an artist or album
-     * in a two hour period may be broadcast. Any more is a breach of this licence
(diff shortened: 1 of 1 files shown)


src/Classes/ServiceAPI/MyRadio_TracklistItem.php
```diff
@@ -404,7 +405,7 @@ public static function getIfAlbumArtistCompliant(MyRadio_Track $track, $include_
             }
         }
 
-        return $result[0] < 2;
+        return $result[0] = 0;
```
LordAro: eep.

Eiron: <1 maybe? All I know is that it should only return a result if there are no matches within an hour.

Eiron: Oh. ==0. I code good.

LordAro:
Note: Merging this will require pulling myradio on dolby (unless someone gets around to removing that dependency before then)

Eiron:
The error that prompted this change may still be lurking somewhere, but this will ensure we are sticking more stringently to the terms of the license. It will certainly be more obvious if something is not working right with this code in play.

PR #704, opened 2017-06-12
$t->getArtist() == $track->getArtist()

Eiron:
Changed the "===" to match the getAlbum format above, as the getArtist check was still not working. Not that I know much about how anything works, but maybe this will fix it?


LordAro:
I don't believe this is the right action here, and certainly wouldn't fix any artist check, breaking it if anything at all.

If anything, the other "==" should be an "===" to avoid type coercion

Eiron:
The album check definitely is working, though (or at least it appears to be). It is the artist check that is failing. There is something wrong with how the check is running.

PR #712, opened 2017-10-28
Give unscheduled seasons a timestamp, not string.

mstratford:
Because it should be outputting a date string all the time, not arbitrary "Not Scheduled".


mstratford:
I've realised that this is going to directly affect the UI of myradio in the scheduler because eww.

LordAro:
handle it in the controller, not the class!

mstratford:
Wondering thoughts on this?

PR #716, opened 2017-12-09
Implement Twitter Mentions into SIS Messages

### LINKED ISSUE #347: Support for twitter in SIS
Integration with the Twitter API to see hashtags/mentions and possibly reply to mentions.

From T36

TomBellis:
An attempt to resolve #347 

Any mentions of URY1350 on twitter during a show would appear in SIS Messages.

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Timeslot.php
+++ b/src/Classes/ServiceAPI/MyRadio_Timeslot.php
@@ -1044,6 +1044,47 @@ class MyRadio_Timeslot extends MyRadio_Metadata_Common
     }
 
     /**
+     * Sends a tweet to the timeslot for display in SIS.
+     *
+     * @param  string $tweet the tweet(mention) to be sent
+     * @param  string $sender the sender of the tweet to be sent
+     * @return MyRadio_Timeslot
+     */
+    public function sendTweet($tweet, $sender)
+    {
+        $tweet = trim($tweet);
+
+        if (empty($tweet)) {
+            throw new MyRadioException('Tweet is empty.', 400);
+        }
+
+        $junk = SIS_Utils::checkMessageSpam($tweet);
+        $warning = SIS_Utils::checkMessageSocialEngineering($tweet);
+
+        if ($warning !== false) {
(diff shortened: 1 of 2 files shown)


TomBellis:
Would quite like it to look something like this...

screen shot 2017-12-09 at 23 34 09

TomBellis:
@LloydW93 has pointed out that tweetID would be longer than comm_source character varying(15), so need to reconsider that.

adfw:
Could simply update the database table -- AFAIK there's no particular reason that field couldn't become wider.

TomBellis:
@adfw That might make sense. Twitter say tweetID is an Int64 although they currently only use 18 digits. 
https://developer.twitter.com/en/docs/tweets/data-dictionary/overview/tweet-object

LordAro:
Quite a lot of duplication between sendMessage & sendTweet (which is also poorly named? Not like it actually sends the tweet)

I'm not clear when/how sendTweet actually gets called either, is there something missing from the diff, or is there supposed to be something else running in the background? If the latter, it would probably be better as a Daemon rather than an external service

TomBellis:
@LordAro This is just the start I created the pull request so I could ask for input. 

I’ve not yet added the part which gets the tweets from twitter’s API. I was thinking it would be a deamon which calls the “sendTweet” to add the details of the tweet to the messages table? 

There’s quite a lot of duplication because sendMessage and sendTweet do very similar things, but sendTweet would take more parameters (message, name (screen name), tweetID). If there is a better way to do this I am happy to look into it.

LordAro:
Generalised function that takes some message and some id would do the trick, would it not?

Brookke:
We could add a sender field and maybe source field (URL to the tweet or “WEB” or “TEXT”) to the current sendMessage method. Then we could add a sender (from) box to the site. And the same api method could be used for the twitter flow.

It would also be nice if we could click through to the tweet.

Brookke:
Just an update on this for anyone who looks at it and decides they want to continue it.

Above comments make the most sense making a generic message function. Using this some sort of separate service would need to be written that checks for mentions sent to our twitter account and then calls the send message api method. Some duplication checking via id maybe need to be considered

PR #718, opened 2017-12-14
Make API say that we're Off Air when we are.

mstratford:


src/Classes/ServiceAPI/MyRadio_Scheduler.php
```diff
@@ -143,6 +150,36 @@ public static function getTerm($termid)
 
         throw new MyRadioException('That term could not be found', 400);
     }
+    
+    /**
+     * Returns the current term, if there is one.
+     *
+     * @todo There's currently no caching on this, could be a potential slowdown
+     *
+     * @return Array of a term
```
LordAro: array of a term? an array with a single value seems weird...

src/Classes/ServiceAPI/MyRadio_Scheduler.php
```diff
@@ -131,7 +131,14 @@ public static function getTerms()
             ORDER BY start ASC'
         );
     }
-
+
```
LordAro: muh trailing

src/Classes/ServiceAPI/MyRadio_Scheduler.php
```diff
@@ -131,7 +131,14 @@ public static function getTerms()
             ORDER BY start ASC'
         );
     }
-
+    
+    /**
+     * Returns term item of the id $termid.
+     *
+     * @todo There's currently no caching on this, could be a potential slowdown
```
LordAro: caching? it's a single, simple sql query, fairly sure no caching is required here. Database has bigger problems if this is an issue

src/Classes/ServiceAPI/MyRadio_Scheduler.php
```diff
@@ -143,6 +150,36 @@ public static function getTerm($termid)
...
+    
+    /**
+     * Returns the current term, if there is one.
+     *
+     * @todo There's currently no caching on this, could be a potential slowdown
+     *
+     * @return Array of a term
+     */
+    
+    public static function getActiveTerm()
```
LordAro: duplicates almost all of getTerms. how much would it break to pull endtime out of getTerms and just use that?

src/Classes/ServiceAPI/MyRadio_Scheduler.php
```diff
@@ -143,6 +150,36 @@ public static function getTerm($termid)
...
+    }
+    
+    /**
+     * Returns if we are currently in term time.
+     *
+     * @todo There's currently no caching on this, could be a potential slowdown
+     *
+     * @return Boolean
+     */
+    public static function isActiveTerm()
```
LordAro: "isActiveTerm" might suggest that you're querying a particular term or something. Maybe `isInActiveTerm` or just `inTerm` would do?

src/Classes/ServiceAPI/MyRadio_Timeslot.php
```diff
@@ -594,15 +594,29 @@ public static function getWeekSchedule($weekno, $year = null)
...
+                    'end_time' => $next ? $next->getStartTime() : 'The End of Time',
+                ],
+            ];
+        } elseif (empty($timeslot)) {
             //There's currently not a show on.
             $response = [
                 'current' => [
                     'title' => Config::$short_name.' Jukebox',
-                    'desc' => 'Non-stop Music',
+                    'desc' => 'There are currently no shows on right now, even our presenters
```
LordAro: any particular reason for this change? looks a bit like a frontend thing

mstratford: Yup, it is, I just bundled it here since I was editing the above bit too.

src/Classes/ServiceAPI/MyRadio_Scheduler.php
mstratford: Huh, my code editor is setup to do that...

src/Classes/ServiceAPI/MyRadio_Scheduler.php
mstratford: Done :)

src/Classes/ServiceAPI/MyRadio_Scheduler.php
mstratford: I changed it to isTerm() to keep the somewhat convention of "is*" when there's a boolean output.

src/Classes/ServiceAPI/MyRadio_Scheduler.php
mstratford: My dodgy copy pasting from above, bad me...

PR #720, opened 2017-12-18
Add a couple of Podcast API features for 2016site

mstratford:
getPreviousTimeslots is used by UniversityRadioYork/2016-site#82 to get the "You just missed" mixcloud shows. Needs work/advice to actually get more than one result back.

Changes to toDataSource allow 2016-site to have a direct file URI to the podcast files.
Todo: Make sure the podcast interface within myradio hides the extra fields. DONE

Goes with UniversityRadioYork/myradio-go#42

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Podcast.php
+++ b/src/Classes/ServiceAPI/MyRadio_Podcast.php
@@ -607,6 +607,8 @@ class MyRadio_Podcast extends MyRadio_Metadata_Common
             'description' => $this->getMeta('description'),
             'status' => $this->getStatus(),
             'time' => $this->getSubmitted(),
+            'file' => $this->getFile(),
+            'webFile' => $this->getWebFile(),
             'photo' => Config::$public_media_uri.'/'.$this->getCover(),
             'editlink' => [
                 'display' => 'icon',
--- a/src/Classes/ServiceAPI/MyRadio_Podcast.php
+++ b/src/Classes/ServiceAPI/MyRadio_Podcast.php
@@ -44,6 +44,13 @@ class MyRadio_Podcast extends MyRadio_Metadata_Common
     private $submitted;
 
     /**
+     * If the Podcast has been suspended.
+     *
+     * @var bool
+     */
+    private $suspended;
+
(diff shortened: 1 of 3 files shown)


src/Classes/ServiceAPI/MyRadio_Podcast.php
```diff
@@ -792,14 +835,21 @@ public function convert()
     /**
      * Returns all Podcasts. Caches for 1h.
      *
+     * @param int $noResults The number of results to return per page. 0 for all podcasts.
+     * @param int $page The page required.
+     *
      * @return Array[MyRadio_Podcast]
      */
-    public static function getAllPodcasts()
+    public static function getAllPodcasts($noResults = 0, $page = 1)
```
LordAro: num_results instead? params tend to be snake_case, and "no" looks a bit like... "no"

src/Classes/ServiceAPI/MyRadio_Timeslot.php
```diff
@@ -435,6 +435,38 @@ public static function getCurrentTimeslot($time = null, $filter = array(1))
...
+    /**
+     * Gets the previous Timeslots before $time, in reverse chronological order.
+     *
+     * @param int $time
+     * @param int $n defines the number of timeslots you want before this time.
+     * @param     $filter defines a filter of show_type ids
+     *
+     * @return MyRadio_Timeslot
+     */
+    public static function getPreviousTimeslots($time = null, $n = 1, $filter = array(1))
```
LordAro: array(1) => [1] ?

src/Classes/ServiceAPI/MyRadio_Timeslot.php
```diff
@@ -435,6 +435,38 @@ public static function getCurrentTimeslot($time = null, $filter = array(1))
...
+    {
+        $filter = '{'.implode(', ', $filter).'}'; // lolphp http://php.net/manual/en/function.pg-query-params.php#71912
```
LordAro: I think this is in a few places in the codebase. Maybe make a static helper function in the database class?

PR #721, opened 2017-12-20
Make RFTM use the correct API json file.

mstratford:
This allows the RTFM swagger API documentation to follow the API it's installed in, instead of defaulting to the production one.


MattWindsor91:
:shipit: (can't see anything wrong with this---if it were PHP I'd suggest using the myradio API path thing in the myradio config, but this isn't)

PR #723, opened 2017-12-20
(NOT READY; PROOF OF CONCEPT) Lightweight schedule endpoint

MattWindsor91:
**This is a proof of concept at the moment.  It's almost certainly full of bugs, not idiomatic, and doesn't mesh well with MyRadio.**

This branch adds in a new API endpoint, `timeslot/getLightSchedule/{start}[?end=...]`, which performs a lightweight schedule query, pulling in a summary of each timeslot that is on air between `start` and `end` inclusive (`end` defaults to `start` plus one week).

Unlike `getWeekSchedule` and `get9DaySchedule`, `getLightSchedule` doesn't retrieve `Timeslots`, but instead does one (and only one) honking big SQL query, and (currently) dumps the results out directly.  Each result contains the most recent active show title and description; the timeslot ID; and the timeslot start, end, and duration.

Why another endpoint that bypasses half of MyRadio?  **Performance!**  The idea here is that things like the 2016-site schedule page don't need all of the `Timeslot` information for every show on the week schedule, but just enough information to render the schedule---and, since we don't need full `Timeslot`s, we can get all of the metadata we need in bulk.  The overhead on `weekSchedule` etc is extremely noticeable when loading up a schedule that isn't in cache, and still hurts when the schedule _is_ in cache.

Things that definitely need work:

- Code style, tests, etc. etc.
- Name
- Input format/parsing for `start` and `end` (the code I've laid down is quite nasty)
- Caching (there isn't any)
- If we can integrate this sort of idea into the existing API instead of making a new endpoint, that's great.
- The query itself might not be optimal (I'm not @adfw, so sadly I don't know much about query optimisation)

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Timeslot.php
+++ b/src/Classes/ServiceAPI/MyRadio_Timeslot.php
@@ -585,6 +585,127 @@ class MyRadio_Timeslot extends MyRadio_Metadata_Common
     }
 
     /**
+     * Parses $t as either an integral UNIX timestamp or string RFC-3339
+     * timestamp.
+     *
+     * @param   int|string $t  The time to parse.
+     * @return  DateTime  The parsed date.
+     * @throws  MyRadioException  If $t is the wrong type.
+     */
+    private static function parseLightScheduleTime($t)
+    {
+        if ($t instanceof \DateTime) {
+            $tdt = $t;
+        } else if (is_string($t)) {
+            $tdt = \DateTime::createFromFormat(\DateTime::RFC3339, $t);
+
+            if (!($tdt instanceof \DateTime)) {
+                // Maybe there's a space between date and time, not T.
+                // Postgres etc. do this.
+
+                $tdt = \DateTime::createFromFormat("Y-m-d H:i:sP", $t);
+            }
+        } else if (is_int($t)) {
+            $tdt = new \DateTime("@$t");
+        } else {
+            throw new MyRadioException('Invalid type for schedule timestamp.');
+        }
+
+        if ($tdt instanceof \DateTime) {
+            return $tdt;
+        }
+
+        throw new MyRadioException('Invalid date format.');
+    }
+
+    /**
+     * Gets a summary of Timeslots airing between `start` and `end` inclusive.
+     *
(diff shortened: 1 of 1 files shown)


Brookke:
My initial thoughts on this is for us to implement some sort of partial response something like 
`GET /timeslot/weekschedule/4?year=2017&fields=title,start_time,duration,timeslot_id`
Although I am not sure if this would actually fix the performance fetching the data in the first place, but it would reduce the amount of data needed to be supplied.

Ideas from http://yaoganglian.com/2013/07/01/partial-response/

LordAro:
I'd expect something like /timeslot/weekschedule/2017/4 for a REST API, but i'm not actually sure myradio can manage such a thing

"fields" would be handled by some form of mixin, i'd expect

MattWindsor91:
@Brookke if you were doing that as a simple filter at the PHP level that'd probably be easy to implement, but if you wanted to stop the database from pulling down those fields too I'm not sure how you'd compute the resulting query without strcatting together bits of SQL.

This PRQ is mainly trying to avoid the performance hit of having to run a query to find timeslot IDs matching the schedule range _then_ waking up each timeslot (which means a separate query for each) by fetching all of the relevant information in a single query.

It doesn't then feed that query data into a bulk wakeup of timeslots because A) timeslot construction involves database queries, so I couldn't see how to make a timeslot without doing redundant queries, and B) the use case for this query doesn't need all of the timeslot information anyway.

@LordAro I _think_ MyRadio can only have at most one URL parameter, inasmuch as I originally tried to keep both start and finish on this one as mandatory parameters and MyRadio just assumed both were query params.

Also, the low-level-ness of this endpoint isn't necessary, I can't think of many use cases for this where you _don't_ want a week schedule (the only ones that make sense to me are trying to pull down per-day schedules, which I think 2016site is going to abolish, and maybe if you wanted 9-day schedules too...

MattWindsor91:
Should this prq be put out of its misery?  I still think something like this should happen, but don’t have any real ability left to champion it myself

mstratford:
At this point, it might be re-implemented using Go for 40K booking stuff...

PR #724, opened 2017-12-26
Redo tests with codecept

LordAro:
Decidedly unfinished, just PRQing for the travis run


LordAro:
Well that seems to work, now to reimplement the existing tests

mstratford:
tests/_support/_generated/ApiTesterActions.php (or more generally) needs to be excluded in gitignore I think :)

LordAro:
That's what tests/_support/_generated/.gitignore is for (which is itself autogenerated)

PR #729, opened 2018-01-16
Change default college to be Unknown

Danzibob:
Noticed this when testing alumni form, clearly incorrect.

tl;dr: this is URY, hence legacy

### DIFF
--- a/src/Classes/Config.php
+++ b/src/Classes/Config.php
@@ -476,9 +476,9 @@ final class Config
 
     /**
      * The default college for new users that do not specify one.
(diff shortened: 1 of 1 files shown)


LloydW93:
This should be changed in URY's MyRadio_Config.local.php - in a new install (e.g. through vagrant), 1 is the correct value.

LordAro:
More things would be broken if this was the case in live site

mstratford:
We've just added a line to the local config to make the fault here fixed.

PR #733, opened 2018-01-25
Fix Uploader not knowing which user uploaded.

mstratford:
Prevents the exception that probably causes uploads to disappear after cache resets.


mstratford:
I don't understand, because I tested the whole upload process and it definitely set the digitised by to myself...

PR #735, opened 2018-02-04
Actually output multiple previous timeslots.

mstratford:


src/Classes/ServiceAPI/MyRadio_Timeslot.php
```diff
@@ -463,7 +463,11 @@ public static function getPreviousTimeslots($time = null, $n = 1, $filter = [1])
         if (empty($result)) {
             return;
         } else {
-            return self::getInstance($result[0]);
+            $output = [];
+            for ($i = 0; $i < sizeof($result); $i++) {
+                $output[$i] = self::getInstance($result[$i]);
```
LloydW93: CoreUtils::resultSetToObjArray?

PR #739, opened 2018-04-01
Migrate Webcams / Selector in SIS to Off Air Source

mstratford:
New Features:
- Ability to switch to off air source (with "are you sure" confirmation).

Fixes:
- Hall & Office webcams now actually selectable.
- More consistent naming between sources and webcams. (Outside vs OB etc).

Tested on MyRadio dev with all webcams now selectable (during a scheduled show).
Selector tested switching from Studio 1 to Jukebox to Off Air.

### DIFF
--- a/src/Public/css/sis.css
+++ b/src/Public/css/sis.css
@@ -89,12 +89,20 @@ fieldset { padding:0; border:0; margin-top:25px; }
 }
 .selbtn {
   height: 35px;
-  width: 35px;
-  background-color: #eee;
+  min-width: 30px;
+  background-color: #ddd;
+  text-align: center;
+  margin: 5px;
+  font-size: 1.2em;
+  padding: 5px 0;
+  color: #fff;
+  display: inline-block;
 }
 .poweredon { cursor:pointer; }
 .s1on, .s2on, .s3on, .s4on, .s8on {
   border: 5px solid gold;
+  height: 25px;
+  box-sizing: content-box;
 }
 .poweredon.s1on { background-color: #f00; }
 .poweredon.s1off { background-color: #600; }
@@ -107,9 +115,7 @@ fieldset { padding:0; border:0; margin-top:25px; }
 .poweredon.s8on { background-color: #ffa500; }
 .poweredon.s8off { background-color: #9c6600; }
 
-
-
-.poweredoff { background-color: #eee; }
+.poweredoff { background-color: #ddd; }
 
 
 .td-msgitem, .delbutton {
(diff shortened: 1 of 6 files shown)


src/Public/css/sis.css
```diff
@@ -85,26 +85,38 @@ fieldset { padding:0; border:0; margin-top:25px; }
...
+  text-align: center;
+  margin: 5px;
+  font-size: 1.2em;
+  padding: 5px 0;
+  color: #fff;
+  display: inline-block;
 }
 .poweredon { cursor:pointer; }
-.s1on, .s2on, .s3on, .s4on {
+.s1on, .s2on, .s3on, .s4on, .s8on {
```
LordAro: something to perhaps consider for future - redo the class/id of this stuff, this one in particular seems perfect for a class, rather than an ID

src/Public/js/sis.selector.js
```diff
@@ -23,15 +25,31 @@ var Selector = function () {
       }
 
       if (locked) {
-        myradio.createDialog("Error", "Could not change studio.<br>Studio selector is currently locked out.");
+        myradio.createDialog("Selector Error", "Could not change studio.<br>Studio Selector is currently locked out.");
         return;
       }
+
+      if (studio == 8) {
```
LordAro: why only a confirm for offair? seems like the sort of thing you should confirm regardless - we've definitely seen instances of people accidentally switching to jukebox before

mstratford: Fair enough, will do :)

mstratford: Done

src/Public/css/sis.css
mstratford: Erm, not quite sure what you mean? Those are classes?

LordAro: I can CSS. In which case, why does each "station" have its own class? Should either be IDs, or more generically named (`powered`, or whatever)

src/Public/js/sis.selector.js
```diff
@@ -23,15 +25,25 @@ var Selector = function () {
       }
 
       if (locked) {
-        myradio.createDialog("Error", "Could not change studio.<br>Studio selector is currently locked out.");
+        myradio.createDialog("Selector Error", "Could not change studio.<br>Studio Selector is currently locked out.");
         return;
       }
+
+      if (confirm == studio) {
+        selStudio(studio); // If they click the same source twice, select it.
```
LordAro: should reset confirm to 0? Should probably do so in the case that confirm != 0, as well. Maybe outside the if else?

mstratford: Ah, woops. My idea was that if you selected the wrong one, you didn't have to cancel it by clicking another source, (confirm=0), but it would just ask you to confirm the second source instead. But yes, confirm=0 should happen when you confirm the selection, since you could potentially get a case where you didn't confirm a quick outside-initiated change.

PR #740, opened 2018-04-01
Get multiple results from previousTimeslots.

mstratford:
Tested on myradio dev (and 2016-site)


src/Classes/ServiceAPI/MyRadio_Timeslot.php
```diff
@@ -460,7 +460,10 @@ public static function getPreviousTimeslots($time = null, $n = 1, $filter = [1])
             [CoreUtils::getTimestamp($time), $n, $filter]
         );
 
-        return self::resultSetToObjArray($result);
+        foreach ($result as $r) {
+            $timeslots[] =  self::getInstance($r['show_season_timeslot_id']);
```
LordAro: not entirely sure how PHP scoping rules work, but probably a good idea to declare this before the foreach loop

(also the extra space after the = on this line)

PR #747, opened 2018-04-23
Only allow jukebox requests to be made if clean.

mstratford:
Fixes UniversityRadioYork/Issues#123

### DIFF
--- a/src/Classes/iTones/iTones_TrackRequest.php
+++ b/src/Classes/iTones/iTones_TrackRequest.php
@@ -93,7 +93,8 @@ class iTones_TrackRequest
      */
     private function trackCanBePlayed()
     {
-        return (!(MyRadio_TracklistItem::getIfPlayedRecently($this->track)) && ($this->track->getClean()));
+        return (!(MyRadio_TracklistItem::getIfPlayedRecently($this->track)) &&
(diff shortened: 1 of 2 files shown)


mstratford:
Tested on Dev, checked a clean song, a song that had recently been played, and an explicit one. Did what it was supposed to :)

src/Classes/iTones/iTones_TrackRequest.php
```diff
@@ -93,7 +93,8 @@ private function canRequestTrack()
      */
     private function trackCanBePlayed()
     {
-        return !(MyRadio_TracklistItem::getIfPlayedRecently($this->track));
+        return (!(MyRadio_TracklistItem::getIfPlayedRecently($this->track)) &&
+               ($this->track->getClean() !== 'n'));
```
LordAro: excessive number of brackets?

` return !MyRadio_TracklistItem::getIfPlayedRecently($this->track) && $this->track->getClean() !== 'n';` ?

mstratford:
It would appear travis thinks that many more lines have changed in this PR than is actually the case?

LordAro:
odd, maybe it's been updated or something. just looks like one file, fix it here anyway?

PR #748, opened 2018-04-23
Re-Implement OBIT to play from Playlist

mstratford:
- [x] Disabling single file played still needs to be done.

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Selector.php
+++ b/src/Classes/ServiceAPI/MyRadio_Selector.php
@@ -404,25 +404,23 @@ class MyRadio_Selector
      * SERIOUSLY, KNOW WHAT YOU ARE DOING WITH THIS METHOD.
      *
      * Calling this method will *terminate station output*, replacing it with
-     * our pre-mixed audio for use in cases of national emergency such as
+     * our OBIT playlist for use in cases of national emergency such as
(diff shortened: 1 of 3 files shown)


mstratford:
The only issue I could see with doing this, is that Jukebox could potentially resort to it's internal default playlists if MyRadio were lost during OBIT time?

adfw:
I'd be inclined to agree with that risk there -- you ideally want it to be as failure-proof as possible. Also, in its current state, would this permit manual track requests to be queued, effectively inserting stuff into the Obit playlist?

mstratford:
Hmm, yes, potentially people could request files, hmm. I could of course disable the request form during OBIT

LordAro:
Aside: Obit is not an acronym - shorthand for obituary

markspolakovs:
As discussed IRL, not only are there the issues above (requests and myradio outages), there's also the risk of playing inappropriate jingles ("WHEN WE'RE ALL OUT LIVING OUR LIVES, HAVING FUN..." isn't great during a royal death). I think the most failure-proof solution is to have Jukebox switch to an "obit mode", where it only plays suitable music (either from MyRadio or a local backup file) and only plays suitable jingles (read as: callsign and nothing else).

markspolakovs:
Old PR, and unworkable for reasons listed above, closing.

Issue #756, opened 2018-06-29
Issues with user creation on local copy.


mstratford:
Noticed a couple of issues with creating users on MyRadio on a local install.

The bulk member add form gives blank page with:`Warning : strip_tags() expects parameter 1 to be string, array given - In /vagrant/src/Classes/MyRadio/MyRadioFormField.php on line 612`

Newly created users can't have their profiles viewed.
`The specified Training Status () does not seem to exist`

mstratford:
Also Warning : file_put_contents(/var/log/myradio/errors.log): failed to open stream: No such file or directory - In /vagrant/src/Classes/MyRadioException.php on line 112

Doesn't auto create a log file.

mstratford:
#778 also

PR #759, opened 2018-09-23
Fix Bulk Add by fixing Text/Email fields in Tabularsets.

mstratford:
Fixes ***Warning : strip_tags() expects parameter 1 to be string, array given - In /usr/local/www/myradio/src/Classes/MyRadio/MyRadioFormField.php on line 612*** errors.


LordAro:
(but you should fix travis)

mstratford:
#756

PR #761, opened 2018-09-25
Limit the length of tags and comma delimit.

### LINKED ISSUE #582: Change tags to be comma delimited

### LINKED ISSUE #751: Limit tag length
https://github.com/UniversityRadioYork/2016-site/issues/125 doesn't display well on the new site, while this could be fixed, having super long specific tags defeats the point of tabs in the first place.

mstratford:
Fixes #582
Fixes #751

2016-site has already implemented shortening via page overflow, but this forces each tag to be 24 chars or less (picked by @Danzibob to be nice looking).

- [x] Podcast changes incoming

### DIFF
--- a/src/Classes/MyRadio/CoreUtils.php
+++ b/src/Classes/MyRadio/CoreUtils.php
@@ -651,7 +651,8 @@ class CoreUtils
         return $exploded_tags;
     }
 
-    public static function checkUploadPostSize(){
+    public static function checkUploadPostSize()
+    {
         // Check that any files don't go over the PHP post_max_size
         // Otherwise, sometimes PHP won't return an error, causing an empty $_POST.
         // This would cause confusing errors relating to empty fields.
--- a/src/Classes/MyRadio/CoreUtils.php
+++ b/src/Classes/MyRadio/CoreUtils.php
@@ -620,6 +620,34 @@ class CoreUtils
         return $needle === '' || strrpos($haystack, $needle, -strlen($haystack)) !== false;
     }
 
+    /**
+     * Explode tags from string to array.
+     * We want to handle the case when people delimit with commas, spaces, or commas and
+     * spaces, as well as handling extended spaces.
+     *
+     * @param string $tags A tags string, comma separated.
+     *
+     * @return array The exploded tags.
+     *
+     * @throws MyRadioException when tags are longer than 24 characters.
+    **/
+    public static function explodeTags($tags)
+    {
+        $tags = preg_split('/[, ] */', $tags, null, PREG_SPLIT_NO_EMPTY);
+        $exploded_tags = [];
+        foreach ($tags as $tag) {
+            if (empty($tag)) {
+                continue;
+            }
+            if (strlen($tag) > 24) {
+                throw new MyRadioException('Sorry, individual tags longer than 24 characters
+                    aren\'t allowed. Please try again.', 400);
+            }
+            $exploded_tags += $tag;
+        }
+        return $exploded_tags;
+    }
+
     private function __construct()
     {
     }
--- a/src/Classes/MyRadio/CoreUtils.php
+++ b/src/Classes/MyRadio/CoreUtils.php
@@ -650,6 +650,37 @@ class CoreUtils
         return $exploded_tags;
     }
 
+    public static function checkUploadPostSize(){
+        // Check that any files don't go over the PHP post_max_size
+        // Otherwise, sometimes PHP won't return an error, causing an empty $_POST.
+        // This would cause confusing errors relating to empty fields.
+        // @see https://stackoverflow.com/questions/2133652/how-to-gracefully-handle-files-that-exceed-phps-post-max-size
+        $post_size = trim(ini_get('post_max_size'));
+        if ($post_size != '') {
+            $last = strtolower(substr($post_size, -1));
+        } else {
+            $last = '';
+        }
+        switch ($last) {
+            // The 'G' modifier is available since PHP 5.1.0
+            case 'g':
+                $post_size *= 1024;
+                // fall through
+            case 'm':
+                $post_size *= 1024;
+                // fall through
+            case 'k':
+                $post_size *= 1024;
+                // fall through
+        }
+        if ($_SERVER['CONTENT_LENGTH'] > $post_size) {
+            throw new MyRadioException(
+                "The content uploaded in this form was too large for the server's configuration.",
+                500
+            );
+        }
+    }
+
     private function __construct()
     {
     }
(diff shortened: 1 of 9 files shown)


src/Classes/ServiceAPI/MyRadio_Season.php
```diff
@@ -257,22 +257,30 @@ public static function create($params = [])
 
         //Same with tags
         if (!empty($params['tags'])) {
-            $tags = explode(' ', $params['tags']);
+            // Explode the tags
+            // We want to handle the case when people delimit with commas, spaces, or commas and
+            // spaces, as well as handling extended spaces.
+            $tags = preg_split('/[, ] */', $params['tags'], null, PREG_SPLIT_NO_EMPTY);
```
LordAro: i'd probably be tempted to split on [, ] and then trim & discard empties. a bit more work, but i think it'll be cleaner code-wise

src/Classes/ServiceAPI/MyRadio_Season.php
```diff
@@ -257,22 +257,30 @@ public static function create($params = [])
...
             foreach ($tags as $tag) {
                 if (empty($tag)) {
                     continue;
+                } else if (strlen($tag) < 25) {
```
LordAro: invert this condition so the error handling is together (and the diff is smaller)

src/Classes/ServiceAPI/MyRadio_Show.php
```diff
@@ -292,16 +292,24 @@ public static function create($params = [])
...
+        $tags = preg_split('/[, ] */', $params['tags'], null, PREG_SPLIT_NO_EMPTY);
         foreach ($tags as $tag) {
-            self::$db->query(
-                'INSERT INTO schedule.show_metadata
-                (metadata_key_id, show_id, metadata_value, effective_from, memberid, approvedid)
-                VALUES ($1, $2, $3, NOW(), $4, $4)',
-                [self::getMetadataKey('tag'), $show_id, $tag, $creator],
-                true
-            );
+            if (strlen($tag) < 25) {
```
LordAro: same as above

src/Controllers/Scheduler/editShow.php
```diff
@@ -33,11 +34,18 @@
         $show->setMeta('title', $data['title']);
         $show->setMeta('description', $data['description']);
 
-        // We want to handle the case when people delimit with commas, or commas and
+        // We want to handle the case when people delimit with commas, spaces, or commas and
         // spaces, as well as handling extended spaces.
+        $tags = preg_split('/[, ] */', $data['tags'], null, PREG_SPLIT_NO_EMPTY);
+        foreach ($tags as $tag) {
+            if (strlen($tag) >= 25) {
```
LordAro: in my mind `> 24` looks cleaner than `>= 25`

src/Classes/ServiceAPI/MyRadio_Season.php
mstratford: Halfway there with my solution now.

src/Classes/ServiceAPI/MyRadio_Podcast.php
```diff
@@ -605,7 +611,7 @@ public function setSuspended(bool $is_suspended)
      *
      * @param MyRadio_Show $show
      */
-    public function setShow(MyRadio_Show $show)
+    public function setShow($show)
```
LordAro: Why this change (and subsequent `if ($show instanceof MyRadio_Show) {`) ?

src/Classes/MyRadio/CoreUtils.php
```diff
@@ -620,6 +620,34 @@ public static function startsWith($haystack, $needle)
...
+        $exploded_tags = [];
+        foreach ($tags as $tag) {
+            if (empty($tag)) {
+                continue;
+            }
+            if (strlen($tag) > 24) {
+                throw new MyRadioException('Sorry, individual tags longer than 24 characters
+                    aren\'t allowed. Please try again.', 400);
+            }
+            $exploded_tags += $tag;
```
LordAro: should strip/trim tag as well - something like "foo ,bar" would result in "foo ", "bar"

src/Classes/ServiceAPI/MyRadio_Podcast.php
Brookke: yeah no idea about this @mstratford mind clarifying?

mstratford: Erm, I'm not sure either, hmm

mstratford:
This is looking done now :)

src/Classes/MyRadio/CoreUtils.php
```diff
@@ -620,6 +620,35 @@ public static function startsWith($haystack, $needle)
...
+    **/
+    public static function explodeTags($tags)
+    {
+        $tags = preg_split('/[, ] */', $tags, null, PREG_SPLIT_NO_EMPTY);
+        $exploded_tags = [];
+        foreach ($tags as $tag) {
+            if (empty($tag)) {
+                continue;
+            }
+            if (strlen($tag) > 24) {
```
LordAro: I wonder whether the length check with the exception should be moved outside the function - make it a proper "pure" function that's more easily reused

src/Classes/ServiceAPI/MyRadio_Season.php
```diff
@@ -400,7 +395,7 @@ public function getEditForm()
                 $this->getID(),
                 [
                     'description' => $this->getMeta('description'),
-                    'tags' => implode(' ', $this->getMeta('tag')),
+                    'tags' => implode(',', $this->getMeta('tag')),
```
LordAro: `', '` ?

src/Controllers/Podcast/editPodcast.php
```diff
@@ -17,7 +18,7 @@
         $podcast = MyRadio_Podcast::create(
             $data['title'],
             $data['description'],
-            explode(' ', $data['tags']),
+            $data['tags'],
```
LordAro: i wonder whether it would be nicer if the (various) create functions took a list of tags, rather than an arbitrary string to be split...

src/Classes/ServiceAPI/MyRadio_Season.php
mstratford: True, however given the entire edit season page doesn't work anyway...

src/Classes/MyRadio/CoreUtils.php
mstratford: Not sure what you mean?

src/Classes/ServiceAPI/MyRadio_Podcast.php
mstratford: It would be because $show can be null (if no show has been selected for the podcast), and PHP doesn't treat that as an acceptable MyRadio_Show object.

src/Controllers/Podcast/editPodcast.php
mstratford: Hmm, maybe, personally given they only get used once, i'm not too fussed.

PR #764, opened 2018-09-30
Fix the rest of the potention xss issues in typeahead

Brookke:
Have not fully tested this yet


Brookke:
Have made it live in myradio-dev and have tested

PR #768, opened 2018-10-08
Update url to point to new jingle form

Brookke:


Brookke:
Ok so production have confirmed that this is the correct url. The reason the url has change is because the old one was owned by a member rather than the station account and thus may disappear when they leave etc...

PR #770, opened 2018-10-19
I wanted to fix some issues

LordAro:
As per usual, completely untested and needs properly reviewing & testing

### DIFF
--- a/src/Classes/MyRadioEmail.php
+++ b/src/Classes/MyRadioEmail.php
@@ -23,11 +23,11 @@ class MyRadioEmail extends ServiceAPI
      * @todo Hardcoded URLs
      */
     private static $headers = 'Content-type: text/plain; charset=utf-8';
-    private static $sender = 'From: MyRadio <no-reply@ury.org.uk>';
+    private static $sender = 'From: MyRadio <no-reply@'.Config::$email_domain.'>';
     private static $footer = 'This email was sent automatically from MyRadio. '
-        .'You can opt out of emails by visiting https://ury.org.uk/myradio/Profile/edit/.';
+        .'You can opt out of emails by visiting https:'.Config::$base_url.'Profile/edit/.';
     private static $html_footer = 'This email was sent automatically from MyRadio. '
-        .'You can opt out of emails <a href="https://ury.org.uk/myradio/Profile/edit/">on your profile page</a>.';
+        .'You can opt out of emails <a href="'.Config::$base_url.'Profile/edit/">on your profile page</a>.';
     // Standard
     /**
      * @var string carriage return + newline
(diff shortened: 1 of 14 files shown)


src/Classes/MyRadioEmail.php
```diff
@@ -200,11 +187,15 @@ private function getHeader()
 
     private static function addFooter($message)
     {
-        return $message.self::$rtnl.self::$rtnl.self::$footer;
+        $footer = 'This email was sent automatically from MyRadio. '
+            .'You can opt out of emails by visiting https:'.Config::$base_url.'Profile/edit/.';
```
mstratford: Thoughts on changing these to use the urlutils generator?

src/Controllers/Setup/strings.php
```diff
@@ -67,8 +67,9 @@
         }
     }
     /* Make sure log directory exists as well */
-    if (!is_dir(dirname(Config::$log_file))) {
-        mkdir($dirname(Config::$log_file), 0755);
+    $log_dir = dirname(Config::$log_file);
+    if (!is_dir($log_dir) || !mkdir($log_dir, 0755)) {
+        die("Could not create log directory at '$log_dir'");
```
mstratford: I presume this was one of the changes mentioned? Shouldn't there be less quotes and more concat dot's here?

src/Classes/ServiceAPI/MyRadio_Season.php
```diff
@@ -608,7 +608,8 @@ public function reject($reason, $notify_user = true)
         $this->setMeta('reject-reason', $reason);
 
         if ($notify_user) {
-            $sname = Config::$short_name;
+            $sname = Config::$short_name; // Heredocs can't do static class variables
```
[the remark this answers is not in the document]
LloydW93: http://php.net/manual/en/language.types.string.php > "Heredoc"

src/Controllers/Scheduler/cancelEpisode.php
```diff
@@ -16,7 +16,7 @@
 
     if (!$result) {
         $message = 'Your cancellation request could not be processed at this time. '
-            .'Please contact programming@ury.org.uk instead.';
+            .'Please contact programming@'.Config::$email_domain.' instead.';
```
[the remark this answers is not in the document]
LloydW93: The main reason this condition should get hit is because you're trying to cancel your show too close to TX - so it is indeed a PC thing.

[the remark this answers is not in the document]
mstratford: Hmm, maybe the error message needs to be a tad clearer then :)

src/Classes/MyRadioEmail.php
LordAro: Probably could do, now that they're definitely not static class variables

I'd like to fix it so base_url doesn't expect to have a trailing slash (unlike all the other config domains...), but I'm not sure how much work that would be

src/Controllers/Setup/strings.php
LordAro: That's silly, it should be &&

Brookke:
Looks good just needs testing

mstratford:
I tried psql 10 and 11, it doesn't seem like travis is quite ready yet.

PR #771, opened 2018-11-07
BAPS-style auto start triggers

LloydW93:
Initial support for the BAPS-style drag bars at the bottom of show planner. Might help raise awareness of the feature in BAPS itself. Currently very messy - mixture of JQ and raw JS usage, times don't update correctly whilst dragging, and there's a small jump on the first dragmove event, but it works.

### DIFF
--- a/src/Public/css/planner.css
+++ b/src/Public/css/planner.css
@@ -5,12 +5,12 @@ html {
 body {
   width: 98%;
   min-width: 1200px;
-  height: calc(100% - 42px);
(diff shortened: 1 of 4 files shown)


mstratford:
Oooh, this will be interesting to look at, will try soon!

mstratford:
Sadly I think this is not gonna make it, I'm going to leave it around to try and replicate it in WebStudio/BAPS(N) though :)

PR #772, opened 2018-11-08
Make a few scheduler errors less critical.

mstratford:
Lowers the number of emails Computing gets for common user mistakes.

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Season.php
+++ b/src/Classes/ServiceAPI/MyRadio_Season.php
@@ -888,7 +888,10 @@ EOT
     {
         //Verify that the input time is valid
         if (!isset($params['time']) or !is_numeric($params['time'])) {
-            throw new MyRadioException('No valid Time was sent to the Scheduling Mapper.', 400);
+            throw new MyRadioException(
+                'No valid Time was sent to the Scheduling Mapper.',
+                400
+            );
         }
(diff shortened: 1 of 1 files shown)


src/Classes/ServiceAPI/MyRadio_Season.php
```diff
@@ -888,20 +888,23 @@ public function schedule($params)
...
+            throw new MyRadioException(
+                'No valid Time was sent to the Scheduling Mapper.',
+                400
+            );
         }
         if ($params['time'] != -1 && !isset($this->requested_times[$params['time']])) {
             throw new MyRadioException(
                 'The Time value sent is not a valid Requested Time Reference.',
-                MyRadioException::FATAL
+                500
```
LordAro: Why is this one a 500?

Pretty sure none of these exceptions need to be split over multiple lines either

mstratford: The multiple lines thing was due to 120 char line length.
Could be a 400, wasn't sure given it was once a "FATAL" error.

LordAro: All the others were FATAL as well :p
The only line that's gotten longer is the last one, all the otherse are shorter than they were

mstratford: I think the issue was it was ignoring the lengths, until I changed them and they were still over 120chars. I've changed the 500 now though.

PR #775, opened 2018-11-20
Random quotes

Skwunk:


hicks927:
Php looks fine, apart from the array_map as you've already mentioned.
 The SQL statement... is absolutely perfectly fine... I'd always been taught to do that with 'TOP' rather than limit, though that could be because I learnt queries on an old database that ran on 128MB of RAM and thus was optimised to hell and back in terms of it's caching.
 The '@return array An array of active quote' could probably do with a rewording. Sure it's consistent but it doesn't help anyone who's just getting into MyRadio code <hint: freshers>. But may be consistency is better? Who knows...
 I'll leave it to others for the style sheet stuff, I've never had the skills nor inclination to work that stuff out.

src/Classes/ServiceAPI/MyRadio_Quote.php
```diff
@@ -36,6 +36,16 @@ class MyRadio_Quote extends ServiceAPI
...
 
+    const GET_RANDOM_SQL = '
+        SELECT
+            *
+        FROM
+            people.quote
+        ORDER BY 
+            random()
+        LIMIT 1
+        ;';
```
mstratford: Imho, the first ; should be on the line above

src/Classes/ServiceAPI/MyRadio_Quote.php
```diff
@@ -69,7 +79,7 @@ class MyRadio_Quote extends ServiceAPI
         WHERE
             quote_id = $2
         ;';
-
+
```
mstratford: May want to tell your editor of choice to remove whitespace.

src/Controllers/Quotes/randomQuote.php
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * Allows the viewing of a single random quote
+ */
+use \MyRadio\MyRadio\CoreUtils;
+use \MyRadio\ServiceAPI\MyRadio_Quote;
+
+CoreUtils::getTemplateObject()->setTemplate('table.twig')
+    ->addVariable('title', 'Random Quote')
+    ->addVariable('tabledata', CoreUtils::setToDataSource(MyRadio_Quote::getRandom()))
+    ->addVariable('tablescript', 'myradio.quotes')
```
mstratford: You don't seem to have created the myradio.quotes.js this is referring to :)

src/Menus/Quotes.json
```diff
@@ -10,6 +10,11 @@
                 "title": "Add Quote",
                 "url": "module=Quotes,action=editQuote",
                 "description": "Add a quote to the Quotes Database."
+            },
+            {
+                "title": "Random Quote",
+                "url": "module=Quotes,action=randomQuote",
+                "description": "You feeling lucky!"
```
mstratford: Purely subjective, but shouldn't that be a question mark? :)

src/Controllers/Quotes/randomQuote.php
Skwunk: It uses the same tablescript as the main quotes board page

src/Menus/Quotes.json
Skwunk: Yes it probably should

src/Controllers/Quotes/randomQuote.php
mstratford: D'oh!

Issue #766, opened 2018-10-02
Add ability to delete a training slot.


mstratford:
Effectively like canceling a show, in theory code should be quite similar.

PR #777, opened 2018-11-22
Add ability to cancel demo slots

Brookke:
- users who created the slot
- only if the demo is empty
- also add manual perm to allow people to delete demos

Unfortunately says episodes in some places like the cancelation reason page and the email sent out

fixes: #766


Brookke:
not yet will do that soon and then merge

src/Controllers/Scheduler/listDemos.php
```diff
@@ -16,8 +16,20 @@
...
         $demo['attending'] = MyRadio_Demo::usersAttendingDemo($demo['show_season_timeslot_id']);
+
+        if (MyRadio_Demo::isDemoEmpty($demo['show_season_timeslot_id'])) {
+            $demo['attending'] = 'Empty';
+            $demo = addCancelButton($demo);
+        }
+    } elseif ($currentUser->hasAuth(AUTH_ADDDEMOS)) {
+        $demo['attending'] = MyRadio_Demo::usersAttendingDemo($demo['show_season_timeslot_id']);
+
+        if (MyRadio_Demo::isDemoEmpty($demo['show_season_timeslot_id'] && MyRadio_Demo::getDemoer() == $currentUser)) {
```
mstratford: This line still needs the timeslot ID fed to getDemoer, leads to some very nice errors on dev :P
Probably better to shorten the whole $demo['xxx'] into a var on previous line in that case.

mstratford:
Also, while we're here, mind just touching up the table js to display headings again? :)

mstratford:
Ah, so the actual issue with the display headings, is that the number of columns changes depending on if you're a trainer, or a trainee looking for one, so for people signing up it looks fine. Maybe some work could be done (maybe this is already done when the cancel button is displayed, not sure).

Brookke:
the data tables thing is tricky as you can't set titles by data

Brookke:
Right now only the UI checks to see if the demo is empty, ideally this should also happen in the API. The only way I can see to do that is to duplicate the attendingDemoCount. Any other ideas @mstratford @LordAro?

PR #779, opened 2018-12-02
Added minutes menu item, changed constitution menu link

Skwunk:


LordAro:
You can merge things yourselves, you know :p

Personally, I'd very much recommend limiting the number of external links on the menu page. Constitution & minutes should be published where all members can easily see them, without having to bother with google drive. Of course, that does mean someone needs to fix the wiki auth...

PR #784, opened 2019-02-19
Added includeSuspended parameter to getAll endpoint

Skwunk:
Podcasts always included suspended rows and so broke the website, now it has a parameter.
Changes also made to myradio_go and 2016_site in accordance with this.

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Podcast.php
+++ b/src/Classes/ServiceAPI/MyRadio_Podcast.php
@@ -852,18 +852,23 @@ class MyRadio_Podcast extends MyRadio_Metadata_Common
      *
      * @param int $num_results The number of results to return per page. 0 for all podcasts.
      * @param int $page The page required.
+     * @param bool $includeSuspended Whether to include suspended podcasts in the result
      *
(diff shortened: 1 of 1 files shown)


src/Classes/ServiceAPI/MyRadio_Podcast.php
```diff
@@ -852,18 +852,23 @@ public function convert()
...
+
+        if(!includeSuspended) {
+            $query .= " WHERE suspended = false ";
+        }
 
         $filterLimit = $num_results == 0 ? 'ALL' : $num_results;
         $filterOffset = $num_results * $page;
 
-        $query .= $filterOffset . " LIMIT " . $filterLimit;
+        $query .= "OFFSET " . $filterOffset . " LIMIT " . $filterLimit;
```
LordAro: If includeSuspended is set, the query will be "...ORDER BY submitted DESCOFFSET ..."

Skwunk: Yeah that was a silly mistake should be fixed now

PR #785, opened 2019-02-28
Trying to fix podcasting getAll SQL

Skwunk:
While testing discovered that my last changes didn't work, this might fix it.

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Podcast.php
+++ b/src/Classes/ServiceAPI/MyRadio_Podcast.php
@@ -858,10 +858,11 @@ class MyRadio_Podcast extends MyRadio_Metadata_Common
      */
     public static function getAllPodcasts($num_results = 0, $page = 1, $includeSuspended = true)
     {
-        $where = !$includeSuspended ? 'WHERE suspended = false': '';
+        $where = !$includeSuspended ? 'WHERE suspended = false ': '';
         $filterLimit = $num_results == 0 ? 'ALL' : $num_results;
         $filterOffset = $num_results * $page;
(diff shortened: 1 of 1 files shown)


src/Classes/ServiceAPI/MyRadio_Podcast.php
```diff
@@ -858,17 +858,12 @@ public function convert()
...
     public static function getAllPodcasts($num_results = 0, $page = 1, $includeSuspended = true)
     {
-        $query = "SELECT podcast_id FROM uryplayer.podcast
-                  ORDER BY submitted DESC ";
-
-        if (!includeSuspended) {
-            $query .= "WHERE suspended = false ";
-        }
-
+        $where .= !$includeSuspended ? 'WHERE suspended = false ': '';
```
LordAro: Surprised `.=` on a brand new variable works at all..

src/Classes/ServiceAPI/MyRadio_Podcast.php
```diff
@@ -858,17 +858,12 @@ public function convert()
...
         $filterLimit = $num_results == 0 ? 'ALL' : $num_results;
         $filterOffset = $num_results * $page;
+        $query = "SELECT podcast_id FROM uryplayer.podcast " . $where
```
LordAro: Perhaps use some string interpolation? No need to worry about spaces in the extra bits that way
```suggestion
        $query = "SELECT podcast_id FROM uryplayer.podcast $where ORDER BY submitted DESC OFFSET $filterOffset LIMIT $filterLimit";
```

src/Classes/ServiceAPI/MyRadio_Podcast.php
Skwunk: Removed it anyway, just to be safe.

Issue #767, opened 2018-10-05
Show Planner Importer allows import of expired items.


mstratford:
Still hearing old jingles/banned beds on radio this week :/

mstratford:
Woo, this is done!

PR #786, opened 2019-03-01
Prevent Importing expired/undigitised items into show plans.

mstratford:
Something that should have been there from the start (this has caused lots of people to play old jingles etc on shows because they continuously imported them from show to show).

Tested (and still running) on myradio-dev.

### DIFF
--- a/src/Public/js/nipsweb.import.js
+++ b/src/Public/js/nipsweb.import.js
@@ -161,7 +161,7 @@ function loadChannelList() {
           itemid = data.payload[selectedChannelNo][item]["album"].recordid + "-" + data.payload[selectedChannelNo][item].trackid;
         } else {
           cleanStars = "";
-          extraString = "";
+          extraString = "(" + data.payload[selectedChannelNo][item].length + ")";
           itemid = "ManagedDB-" + data.payload[selectedChannelNo][item].managedid;
           expired = data.payload[selectedChannelNo][item].expired;
         }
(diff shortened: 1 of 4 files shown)


src/Public/js/nipsweb.import.js
```diff
@@ -133,45 +134,56 @@ function updateTimeslotList() {
...
-      for (item in data.payload[selectedChannelNo]) {
-        if (data.payload[selectedChannelNo][item].type == "central") {
-          if (!data.payload[selectedChannelNo][item].clean) {
-            cleanStars = "**";
-          } else {
-            cleanStars = "";
-          }
-          extraString = " - " + data.payload[selectedChannelNo][item].artist + " - " + data.payload[selectedChannelNo][item].album.title + " (" + data.payload[selectedChannelNo][item].length + ")";
-          itemid = data.payload[selectedChannelNo][item]["album"].recordid + "-" + data.payload[selectedChannelNo][item].trackid;
+      currentShowPlan = data.payload;
```
LordAro: Global state :(
How feasible would it be to make this a parameter for loadChannelList ?

mstratford: Not sure what to here, apart from reloading the show plan AJAX every time I change channel (which I fixed in this PR from previous behavior)

mstratford:
Fixes: #767

src/Public/js/nipsweb.import.js
Brookke: Without creating all the channels and then dynamically hiding and showing them, I think this is as good as we are going to get

mstratford: Yeah.

PR #791, opened 2019-03-04
Use the git commit hash to version CSS/JS to bust cache.

mstratford:


src/Classes/MyRadio/CoreUtils.php
```diff
@@ -749,4 +749,17 @@ public static function newPassword()
...
         'Broadcasting',
     ];
+
+    /**
+     * Get the hash of the current git HEAD, used to append to CSS urls for caching
+     * @return mixed Either the hash or a boolean false
+     */
+    public static function getVersion() {
+        $branch_ref = trim(explode("ref: ", file_get_contents('../../.git/HEAD'))[1]);
+        if ( $hash = file_get_contents( sprintf( '../../.git/%s', $branch_ref ) ) ) {
```
LordAro: Single =
Also the spacing is weird

LordAro: Also, my god. This is awful in every way I can think of. Can you not?

Brookke: if this is only run once when myradio is initialised it's probably easier to just use the current date.

PR #799, opened 2019-09-30
Removed quotes board temporary

Lagher0:


mstratford:
Looks good :) We can revert it when management say so.

PR #800, opened 2019-10-17
Show Co-Presenters in the short string.

mstratford:
People like to put themselves as co-presenters. To the normal user, these are also presenters. Show them on RadioPlayer etc.


mstratford:
Warning: This will require the new config option adding to the MyRadio config file on deployment. Values [1,10] match Presenters and Co-Presenters.

mstratford:
RIP, the DB actually has a table variable for this. I'll update this to actually use it instead.

mstratford:
Not sure how to do this without there being yet another DB call.

mstratford:
Closed in favour of #924

PR #801, opened 2019-10-17
Let API get available genres.

mstratford:
Just a thing sitting in my local git. Useful maybe for website to search by Genre...

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Scheduler.php
+++ b/src/Classes/ServiceAPI/MyRadio_Scheduler.php
@@ -194,13 +194,17 @@ class MyRadio_Scheduler extends ServiceAPI
     }
 
     /**
-     * Returns a list of potential genres, organised so they can be used as a SELECT MyRadioFormField data source.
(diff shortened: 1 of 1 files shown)


src/Classes/ServiceAPI/MyRadio_Scheduler.php
```diff
@@ -194,13 +194,17 @@ public static function getTermStartDate($term_id = null)
...
      */
-    public static function getGenres()
+    public static function getGenres($id_only = True)
     {
         self::wakeup();
 
-        return self::$db->fetchAll('SELECT genre_id AS value, name AS text FROM schedule.genre ORDER BY name ASC');
+        $select = $id_only ? "genre_id" : "*";
+
+        return self::$db->fetchAll('SELECT '.$select.' AS value, name AS text FROM schedule.genre ORDER BY name ASC');
```
LordAro: Could use string interpolation instead of concatenation - looks slightly cleaner

AJDurant:
also, SQL injection? Even if $select is always written internally, doing
concatenation for SQL statements is a bad practice.

mstratford:
Turns out I was being silly. It already gives the id's and names.

PR #802, opened 2019-10-18
Upgrade Vagrant to Ubuntu 19.04 / psql11

mstratford:
Also fixes the fact MyRadio can't configure it's own config file on Setup.
It also reloads apache2 on vagrant up, it seems to not load the config correctly otherwise.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,4 +1,4 @@
 *.min.js
-src/vendor/**/*.js
+src/vendor/*
 src/Public/js/vendor/**/*.js
 src/PublicAPI/rtfm/*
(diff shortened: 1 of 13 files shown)


LordAro:
Would be good to sort the remaining CI issues while you're at it

LordAro:
Went ahead and did it anyway :)

mstratford:
Ok, tested this now on MacOS. Would like if someone else could check a new install on a different platform.

markspolakovs:
@mstratford Tested on Ubuntu, seems to work fine

mstratford:
Nice, well, I guess I can't approve it... so :)

PR #803, opened 2019-10-21
Disallow non-upload-trained people from accessing the old uploader

markspolakovs:


markspolakovs:
@mstratford some code uses the method I use, some uses MyRadioExceptions - I presume the latter is the "better" way that new stuff should use?

mstratford:
I switched MyRadio-dev to this, seems to work. Works interesting on mobile (not that MyRadio works too well on mobile anyway) since it's effectively an embed of a page

markspolakovs:
Deployed to myradio-dev, seems fine. Will merge in and deploy to prod as soon as the branch protection gets sorted out.

PR #805, opened 2019-10-25
WIP: moving show colours into MyRadio: Part 1 - Database Schema

markspolakovs:
This is a big project, comprised of several PRs. See [here](https://github.com/orgs/UniversityRadioYork/teams/officers/discussions/1) for the plan (internal only).

This PR sets up the database schema for subtypes.

Design decisions:
- It does not make sense for shows to have multiple subtypes (what colour would it have on the front page then?).
- Neither does it make sense for a show's subtype to change between seasons - the core idea of a show shouldn't change, I think.
- Therefore, we're not using metadata.

Comments welcome.


markspolakovs:
crud, forgot about foreign keys.

adfw:
This looks like a really good idea :)
One comment - there is a (very little used) 'genre' table in the scheduler schema - perhaps this could also be used to auto-colour things? It looks very similar to the 'subtypes' you've already created - so it could work well alongside?

P.S. it's really nice to see someone taking an interest in the database

markspolakovs:
On second thought, it may be worthwhile using a similar structure to `scheduler.show_genre` (with the usual `effective_from` and `effective_to` fields) - a show's subtype **might** actually change (e.g. might get moved to primetime) - although this would make it a slight UI nightmare (show genres don't have a GUI to set `effective_from` either, it's just one field). More thonk necessary.

markspolakovs:
Or maybe we can just tie it into the metadata system? It's a bit of a fustercluck to work with, but it does its job well.

adfw:
Genre with from/to fields sounds appropriate - tie it to show_season possibly ? (show_season_genre). Alternatively yes, the metadata system could work too. For an example of what happens when the genre/colour gets out of kilter, take a look at the schedule from, say 10 years ago - the matching rules sometimes match against things they clearly shouldn't.

MattWindsor91:
One thing I would say is that I’m not sure the colours themselves should be in the database; instead, for separation of presentation and content (and frontend development simplcicity, and general things-being-together-in-the-same-place) reasons, these should remain in the CSS, and the subtype to colour mapping should be done eg by applying subtype CSS classes.

LordAro:
Oh hey, someone's noticed the database migration scripts! You'll probably want to make sure they work before trying to use them...

(myury -> myradio is a change that's been sitting around for nearly 5 years now, would be good to actually get it done, and would be a good testcase)

markspolakovs:
@MattWindsor91 good point - also would make migration on the 2016-site side of things easier. Updated.

@LordAro ha, I'd probably end up running this via psql anyway (don't want to fiddle with the migrator). The myury->myradio migration would need not-insignificant changes on the PHP side, no? That's why I disabled it - don't want to worry about it as part of this PR.

mstratford:
Oooh, looks promising :) Maybe we could merge in master and have builds working again here :)

markspolakovs:
@mstratford Synced with master - the failing build is actually from #803 :grimacing:

LordAro:
https://xkcd.com/1296/

Please try to avoid this, unless you're planning on rebasing/squashing everything together later

See also https://chris.beams.io/posts/git-commit/

markspolakovs:
@LordAro lol, sorry - tbh, makes sense to do a squash, there's a reason I've split this mini-project across multiple PRs

markspolakovs:
Tested the SQL patch against a localhost install, and it seems to apply cleanly, so unless anyone screams I'll squash this in and move on to Part 2.

LordAro:
I'd say that if you're not actually going to use (and fix) the patching mechanism, just apply the changes directly to base.sql
(which should be being done anyway)

markspolakovs:
@LordAro is base.sql idempotent? I get the feeling it isn't. So it might make sense to run this patch on its own in prod, and then and only then merge it into base.sql

LordAro:
base.sql is what is used for new installs. Last time i checked (admittedly a few years ago), it matched what was running in production. It certainly should match, anyway

This is how the upgrade mechanism was supposed to work - base.sql contains the "current" schema, with a particular version number. If the schema version of the currently running database is lower than the source (it's in the PHP somewhere), it runs the relevant upgrade scripts.

markspolakovs:
Right, but I'd need to run my patch on prod before I can add it to base.sql, because I don't want to re-run all of base.sql in prod, right? Also, does the upgrader work on existing installs as well? (To be completely honest, I don't really want to be spending my time troubleshooting the upgrader when I could just run the SQL file and move on.)

LordAro:
Yes, this is the reason myradio has never had any sort of automatic deployment through the CI - because of the potential need for manual database upgrades. Basically what's been done before is that you do it all at once - run the upgrade script yourself (after having taken a backup!), then upgrade the code. The downtime is hopefully limited to a few seconds...

markspolakovs:
I've now applied this in prod, so closing this PR in favour of to #818. I'll dump this in base.sql later.

PR #806, opened 2019-10-28
Prevent broken email when browser wakes.

mstratford:
No longer will we get a wall of HTML over email every time someone leaves the webcams page open and clicks "Report" on the Error popup model. Also was caused by having multiple webcam pages open.


mstratford:
Tested, seems to provide a somewhat decent error message to user name, and doesn't send us an email.

mstratford:
Tested on Dev, no longer throws errors due to 400. Now returns the time counter instead.

PR #807, opened 2019-10-28
Show message if no webcams setup.

### LINKED ISSUE #80: Notice if there are no webcams set up

mstratford:
Fixes #80

### DIFF
--- a/src/Public/js/myradio.webcam.tracker.js
+++ b/src/Public/js/myradio.webcam.tracker.js
@@ -47,6 +47,7 @@ function webcamTrackViewer() {
 
 $(document).ready(
   function () {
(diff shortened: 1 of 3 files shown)


markspolakovs:
Looks good, although can non-comp people see the webcams? Because if they can I'd change the wording to "if you're Computing check the database, if not ..."

mstratford:
The message is aimed at computing, since it will only really appear on a fresh install with permissions set.

markspolakovs:
That phpcs break is my fault, I promise I'll fix it...

mstratford:
I've tested it on a new Vagrant box :)

PR #808, opened 2019-10-28
Add second email to send member their account details

mstratford:
Such that the original welcome email from SM is more personal and can be replied to.

Warning, this will require a config update to add the new email contents and update the existing welcome email.


markspolakovs:
@mstratford code looks good on its face - can you rebase on top of current master so I can pull it in on dev?

mstratford:
Will look at it soon

src/Classes/Config.php
```diff
@@ -673,13 +673,24 @@ final class Config
...
+    /**** SIGNUP EMAILS ****/
+    public static $welcome_email_sender_memberid = null;
     public static $welcome_email = <<<EOT
 
-   This is a welcome email. You can use #NAME to get the user's first name,
-   and include #USER and #PASS to tell them their newly created login details.
+   This is a welcome email. You can use #NAME to get the user's first name.
+
+EOT;
+    public static $account_email = <<<EOT
```
LordAro: this is very generically named, and the documentation below isn't much better. which account? which email? who gives the email to whom?

mstratford: Better? :)

PR #809, opened 2019-10-28
Allow emails to work if the user has URY email without server account.

mstratford:
Because having a user like computing@ury.org.uk does not allow myradio to email them.

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_User.php
+++ b/src/Classes/ServiceAPI/MyRadio_User.php
@@ -499,7 +499,13 @@ class MyRadio_User extends ServiceAPI implements APICaller
                 //ffs, some people don't have an eduroam either.
                 $eduroam = $this->getEduroam();
                 if (empty($eduroam)) {
-                    return;
+                    if (!empty($this->email)) {
+                        //This is an email not associated to a user.
+                        return $this->email;
+                    } else {
+                        //Give up trying to send.
+                        return;
+                    }
                 } else {
                     return $eduroam.'@'.Config::$eduroam_domain;
                 }


LordAro:
The real (other?) issue here is that someone has managed to sign up for a myradio account using a distribution list email. The sign-up should have errored and been disallowed. Then, this particular edge case disappears

mstratford:
@LordAro Well, I guess it's our alumni form that allowed this, since it doesn't stipulate a @york.ac.uk. I was trying to test a new "welcome to URY" email format.

src/Classes/ServiceAPI/MyRadio_User.php
```diff
@@ -499,7 +499,13 @@ public function getEmail()
                 //ffs, some people don't have an eduroam either.
                 $eduroam = $this->getEduroam();
                 if (empty($eduroam)) {
-                    return;
+                    if (!empty($this->email)) {
+                        //This is an email not associated to a user.
+                        return $this->email;
+                    } else {
+                        //Give up trying to send.
+                        return;
```
markspolakovs: This seems like a footgun to me - are we sure that all code that calls `$user->getEmail()` can handle the empty case? AFAICT no other branch returns nothing so I'm slightly concerned here

mstratford: Yup, this is how I found this bug in the first place. It tries to send an email with <> as the email address, so just fails and sticks in the MyRadio email queue.

markspolakovs: Are there any cases in which the new code could cause that failure mode as well?

PR #810, opened 2019-10-29
Update messaging on podcast form fields

markspolakovs:
- clarify that image sizes should be 400x400px
- ask to notify Comp about cache breakage

Context (URY members only): https://ury.slack.com/archives/C091ZRR63/p1572298733002700


mstratford:
So, the requirement is for square images, maybe at least 400x400 instead?
As for the second issue, that should be something that's just fixed in MyRadio itself. I expect it's something in the daemon we need to look at.

markspolakovs:
@mstratford 

First point - ok, updated

Second point - it should, but it'll happen at "the usual URY computing speeds", so let's keep this as a stop-gap, just so we can proactively fix the site rather than someone noticing it's died again

PR #812, opened 2019-11-05
Fix Non well formed numerical value error.

mstratford:
Shows on https://ury.org.uk/myradio/Library/search/ after searching.

### DIFF
--- a/src/Classes/MyRadio/CoreUtils.php
+++ b/src/Classes/MyRadio/CoreUtils.php
@@ -669,6 +669,8 @@ class CoreUtils
         } else {
             $last = '';
         }
+        $post_size = substr($post_size, 0, -1); // Remove the letter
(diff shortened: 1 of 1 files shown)


src/Classes/MyRadio/CoreUtils.php
```diff
@@ -669,6 +669,8 @@ public static function checkUploadPostSize()
         } else {
             $last = '';
         }
+        $post_size = substr($post_size, 0, -1); // Remove the letter
```
LordAro: This could cause issues if $post_size doesn't have a suffix and is purely numeric. Isn't the intval call below enough? string to int casts are normally fine with ignoring trailing characters

mstratford: Tested this, it does indeed :)

PR #813, opened 2019-11-09
Limit SIS message length to 1k chars and fix logic error

markspolakovs:
No reasonable human is going to send more than a tweet's length anyway

### DIFF
--- a/src/Classes/SIS/SIS_Utils.php
+++ b/src/Classes/SIS/SIS_Utils.php
@@ -117,7 +117,7 @@ class SIS_Utils extends ServiceAPI
     public static function checkMessageSpam($message)
     {
         if (strlen($message) > 1000) {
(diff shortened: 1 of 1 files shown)


src/Classes/SIS/SIS_Utils.php
```diff
@@ -116,14 +116,16 @@ public static function readPolls($modules)
      */
     public static function checkMessageSpam($message)
     {
+        if (strlen($message) > 1000) {
+         return true;
```
mstratford: apart from the indentation here, looks good :)

mstratford:
Think you're still missing an extra whitespace on that line at the end. Sorry :P

markspolakovs:
This is why we don't write code in the github online editor.

PR #814, opened 2019-11-10
Lock the showplanner if saving fails

markspolakovs:
If a save fails, showplanner stops saving changes, but still lets users drag tracks around, which could potentially lead to them making changes that aren't saved. To ensure this doesn't happen, disable showplanner prior to reloading.


mstratford:
Mmm, so live atm is with debug enabled to prevent the reload. Might be worth actually changing the warning to explain to reload manually too.

Issue #797, opened 2019-06-13
Link to presenter contract on 'Forbidden' page for show planner


thomasslee97:
If the presenter contract has not been signed, provide a link to the presenter contract when the user tries to access the show planner.

PR #821, opened 2019-11-15
Adding Link to Sign Presetners Contract on Forbidden Page for SIS

michael-grace:
Issue #797 
Also improves a bit of formatting


markspolakovs:
You're going to squash this when you merge... right?

src/Views/Errors/403.php
```diff
@@ -3,10 +3,19 @@
  * This View renders a HTTP/1.1 403 Error - for when a <code>AuthUtils::requirePermission()</code> call returns false.
  */
 use \MyRadio\MyRadio\CoreUtils;
+use \MyRadio\MyRadio\URLUtils;
 
 header('HTTP/1.1 403 Forbidden');
 
-CoreUtils::getTemplateObject()->setTemplate('error.twig')
+if ($err_type == 'contract'){
```
markspolakovs: Not sure I'm a fan of special-casing it like this. Can we perhaps have a switch for the various types of error, defaulting to `error.twig`? Or is there a better solution?

michael-grace: Yeah, I wasn't the biggest fan of doing it this way, but it was like 1:30am. 😂 I was thinking afterwards about doing it slightly differently, I'm going to have a look tomorrow.

michael-grace:
might be ugly, but gives us the benefit of not having to find every point a 403 could be called, and still allows us to add custom links later on

PR #823, opened 2019-11-18
Update base.sql with schema changes

markspolakovs:
Also added back patches/1.sql just in case we ever decide to actually do it


markspolakovs:
Closing in favour of #824.

PR #824, opened 2019-11-18
Moving show colours into MyRadio, part 3 - API

markspolakovs:


markspolakovs:
argh, crappidy crap, it took #823 as a base...

markspolakovs:
Tested on dev. Auth might need defuckenating but we'll sort that out when we do the prod deploy.

PR #825, opened 2019-11-18
Update link to constitution

markspolakovs:


mstratford:
But... But... The wiki was supposed to be the thing that held the constitution. I even updated it the other day :/

markspolakovs:
Well Westie told me it's wrong - we can point myradio to the wiki if we want?

mstratford:
Sorry, github emails apparently don't work. Last I knew it was at the wiki, which linked to (with history applied) the current (at the time) Constitution. Should go here: https://ury.org.uk/wiki/Constitution

PR #828, opened 2019-12-02
Hide payments on profile pages

markspolakovs:
YUSU backpages is dead (rip) and people get confused by "paid: £0.00"


markspolakovs:
Tested on dev.

PR #831, opened 2019-12-19
Add endpoints for available playlists

markspolakovs:
Needed for Campus Playout.

Tested on dev.


markspolakovs:
@mstratford nope, getTrackForJukebox calls getPlaylistFromWeights, which already does similar SQL bollockery, I don't see much potential for cleanup.

PR #832, opened 2019-12-21
Add categories to playlists

markspolakovs:


markspolakovs:
@mstratford that's why I renamed it to .old

schema/patches/1.sql
```diff
@@ -1,21 +1,29 @@
...
+    id SERIAL PRIMARY KEY,
+    name TEXT,
+    description TEXT
+);
+
+INSERT INTO jukebox.playlist_categories (id, name, description)
+VALUES (
+    1,
+    'General',
+    '<p>This category is for all playlists that don''t fit the others. They will not be played by Jukebox or Campus Playout.</p>'
```
mstratford: Either i'm reading this wrong, or is that quote not properly escaped? :)

src/Classes/iTones/iTones_Playlist.php
```diff
@@ -447,6 +469,19 @@ public function setDescription($description)
         $this->updateCacheObject();
     }
 
+    /**
+     * Update the category.
+     * @param $category
+     */
+    public function setCategoryById($category) {
```
mstratford: Maybe worth checking that the category is an int here :)

src/Classes/iTones/iTones_Playlist.php
```diff
@@ -531,14 +638,14 @@ public static function getPlaylistsWithTrack(MyRadio_Track $track)
         return self::resultSetToObjArray($result);
     }
 
-    public static function create($title, $description)
+    public static function create($title, $description, $category)
```
mstratford: also check int here plz

src/Classes/iTones/iTones_Playlist.php
```diff
@@ -513,6 +548,78 @@ public static function getPlaylistFromWeights($playlists_to_ignore = [])
         return self::getInstance(CoreUtils::biasedRandom($result));
     }
 
+    /**
+     * Uses weighted playout values to select a random Playlist from a category, returning it.
+     *
+     * Only includes Playlists with a currently running slot, and a Track.
+     *
+     * @param int $categoryId
```
mstratford: Here you have an int defined, to the API may enforce this, but still, would like to validate properly 👍

src/Classes/iTones/iTones_Playlist.php
```diff
@@ -513,6 +548,78 @@ public static function getPlaylistFromWeights($playlists_to_ignore = [])
...
+    public static function getPlaylistOfCategoryFromWeights($categoryId, $playlists_to_ignore = [])
+    {
+        // TODO: this is a straight copy-paste of the above. If we need to do this again,
+        // consider refactoring.
+        self::wakeup();
+
+        $result = self::$db->fetchAll(
+            'SELECT playlists.playlistid AS item, MAX(playlist_availability.weight) AS weight
+                FROM jukebox.playlists, jukebox.playlist_availability, jukebox.playlist_timeslot
+                WHERE playlists.category = $1
```
mstratford: Is this function actually needed? Could we not just put categoryId param with a default of "*" in the regular getPlaylistFromWeights function?

src/Classes/iTones/iTones_Playlist.php
```diff
@@ -513,6 +548,78 @@ public static function getPlaylistFromWeights($playlists_to_ignore = [])
...
+
+        return self::getInstance(CoreUtils::biasedRandom($result));
+    }
+
+    /**
+     * Find all playlists with a given playlist category.
+     * @param $categoryId
+     * @return iTones_Playlist[]
+     */
+    public static function getAllPlaylistsOfCategory($categoryId)
```
mstratford: guess what :P

mstratford:
Also Travis plz :)

schema/patches/1.sql
markspolakovs: Nope, that's how postgres does quote escapes. Yeah it's ugly.

src/Classes/iTones/iTones_Playlist.php
markspolakovs: That's a bit too close to SQL injection for my liking: we'd have to dynamically construct the SQL if the param is not present - somewhat tame as far as things go, but I'm not wholly convinced of the need. If for some reason we add a third function that does a substantially similar query, then we can look at this again.

markspolakovs:
Remind me, when I get around to my "unbreak travis" PR, to move phpcs after tests, so a code style violation doesn't mask a failing test which is more serious

src/Classes/iTones/iTones_Playlist.php
mstratford: I mean, like you can do funcname([...], $categoryId == "*"), meaning if it's not provided, it's "*". Then, if the value isn't === "*", you check if it's an integer, before putting it into the SQL query. Because afaict the only difference between this and the other function is the "WHERE playlists.category = $1" bit. You can just do "WHERE playlists.category = *" in this case.

markspolakovs: Sadly, that's not a thing you can do.

image

markspolakovs:
@mstratford phpcs is done screaming - the test is still br-oken but that's outside the scope of this PR.

src/Classes/iTones/iTones_Playlist.php
mstratford: aww.

PR #837, opened 2020-01-18
Add debug logging to getTrackForJukebox

markspolakovs:


src/Classes/iTones/iTones_Utils.php
```diff
@@ -357,4 +375,13 @@ public static function telnetEnd()
             self::$telnet_handle = null;
         }
     }
+
+    private static function log(...$fields)
```
mstratford: Ideally this would have a config var for turning it on / off and a logging location.

mstratford:
Do we still want / need this?

markspolakovs:
Nah. Bin.

markspolakovs:
(I'll do a proper logging PR at some point, but not right now.)

PR #839, opened 2020-01-26
Validate Swagger as part of Travis build

markspolakovs:
Note: this will cause build failures - that's because our swagger is invalid as it is. Hopefully this will help us track down issues and make it valid.


LordAro:
Much better to fix existing issues before merging, otherwise no one will ever bother to fix it

PR #840, opened 2020-01-26
Add "nonav" URL parameter

markspolakovs:
Need to make MyRadio^2 work nice.


markspolakovs:
Not intended to be user-visible anyway. Sod it.

LloydW93:
I'll point out I did misread this as non-av though ;)

PR #841, opened 2020-01-31
Bodge the fucking API into submission

markspolakovs:
I will be entirely honest, I have no idea why this works, but it works.

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Swagger2.php
+++ b/src/Classes/ServiceAPI/MyRadio_Swagger2.php
@@ -5,6 +5,7 @@
  */
 namespace MyRadio\ServiceAPI;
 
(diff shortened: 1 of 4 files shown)


mstratford:
Probably not an issue, the output of /user/currentuser is correct, but noticing in the syslog that if you set 'api_key=' (blank), it goes to run getCurrentUserWithoutMessingWithSession() but still returns null. The code above decides that you want session user, but this new code is not.

markspolakovs:
Ah, probably fixed by using `!empty()` instead of `!isset()`. Will update.

markspolakovs:
Build failing, but to be quite honest I have no idea what it's on about - the lines it's complaining about are less than 100 chars..

PR #844, opened 2020-02-29
Add MyRadio^2 Sign-in Handoff

markspolakovs:


mstratford:
Sorry, i've managed to be out of the loop of exactly what this is for. Looks ok, but i'd like to actually see it :P

markspolakovs:
Sorry, my fault for inadequate docs.

In MyRadio^2, there is (or at least, will be) a line along the lines of

```html
<iframe src="https://ury.org.uk/myradio-dev/MyRadio/login/?next=%2fmyradio-dev%2fMyRadio%2fmyr2Handoff%3fnonav%3dtrue&nonav=true" />
```

That iframes the login screen, and redirects to this new page on success. It then sends a message out of the iframe, which prompts mry^2 to check again for sign-in.

PR #845, opened 2020-03-01
Make jukebox requests use replay_gain handler.

mstratford:


markspolakovs:
Hm, I'm not entirely sure we should be doing this from MyRadio, as that seems like a separation-of-concerns snafu.

mstratford:
Maybe, but it is MyRadio just telnetting directly in... So at that point you'd have to change it somehow from within the depths of Liquidsoap queues.

markspolakovs:
If it were me doing this now, I'd do something like what I did for Campus Playout - define a custom function that does all the replaygain fuckery it needs to. Meh, that would require extensive _*LIQUIDSOAP FUCKERY*_ so I'll begrudgingly approve this for now.

PR #846, opened 2020-03-01
Add mixcloud start/end custom times to timeslot.

mstratford:
Add metadata for custom mixclouder upload start/end times.


mstratford:
Yeah, when not set, api shows null, Mixclouder does the usual. If you define a start point, it will run from then to the usual scheduled end. Also works the other way with end point. Will commit those changes too to the mixclouder repo.

mstratford:
(also means you can have the show extended past the beginning and end of the show if needed)

PR #847, opened 2020-03-01
Allow for suspending quotes

mstratford:


markspolakovs:
Reverted this after merge. It should not have passed code review.

PR #850, opened 2020-03-02
Fix sso_session duplications.

mstratford:
Make SQL deal with the race condition of multiple simultaneous new session creations.

### DIFF
--- a/schema/base.sql
+++ b/schema/base.sql
@@ -3733,7 +3733,7 @@ ALTER TABLE ONLY selector
 --
 
 ALTER TABLE ONLY sso_session
-    ADD CONSTRAINT sso_session_pkey PRIMARY KEY (id, "timestamp");
+    ADD CONSTRAINT sso_session_pkey PRIMARY KEY (id);
 
 
 --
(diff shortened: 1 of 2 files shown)


mstratford:
Tested not to break everything else on MyRadio-staging.

schema/base.sql
```diff
@@ -3733,7 +3733,7 @@ ALTER TABLE ONLY selector
 --
 
 ALTER TABLE ONLY sso_session
-    ADD CONSTRAINT sso_session_pkey PRIMARY KEY (id, "timestamp");
+    ADD CONSTRAINT sso_session_pkey PRIMARY KEY (id);
```
markspolakovs: I guess it doesn't matter now that it's deployed to staging, but personally I prefer individual patch files using ALTER (let's ignore the fact that the patcher is bro-ken for now...) instead of modifying base.sql, for ease of deployment

mstratford: hmm. Fair.

markspolakovs: We really should get around to fixing the patcher - perhaps we can finally actually deploy the myury->myradio patch :p

mstratford: I'll open a ticket :)

mstratford:
Can we merge this please? Make the emails staap :P

PR #853, opened 2020-03-02
Fix 404 on null default sender for welcome email.

mstratford:


mstratford:
Travis confirmed fixed.

PR #854, opened 2020-03-04
Attempt 573421934 at fixing session ID races

markspolakovs:


src/Classes/MyRadio/MyRadioSession.php
```diff
@@ -32,6 +32,25 @@ public function open($save_path, $sesion_name)
...
+        $default = new \SessionHandler();
+        while (true) {
+            $testSid = $default->create_sid();
+            $result = $this->db->query('
+                INSERT INTO public.sso_session (id, data, timestamp) 
+                VALUES ($1, \'\', NOW())
+                ON CONFLICT DO NOTHING
+            ', [$testSid]);
+            if ($this->db->numRows($result) > 0) {
+                return $testSid;
```
markspolakovs: In theory this can cause an ininite loop, but I think it's unlikely that it takes more than 2 attempts to generate a SID

markspolakovs:
DO NOT MERGE - bro-ken

PR #860, opened 2020-03-29
Add WebStudio API User Session and Tracklisting

mstratford:
Allows users (with new permissions) to tracklist for shows using WebStudio.

Also allows WebStudio to get the current logged in user/selected timeslot data.

### DIFF
--- a/schema/data-auth.json
+++ b/schema/data-auth.json
@@ -78,7 +78,7 @@
     ["Stop Broadcast","AUTH_STOPBROADCAST", true],
     ["Add Quotes","AUTH_ADD_QUOTES", true],
     ["Upload Music - Manual","AUTH_UPLOADMUSICMANUAL", true]
-    ["User can tracklist own/current show","AUTH_TRACKLIST_OWN", true],
-    ["User can tracklist for any show/timeslot","AUTH_TRACKLIST_ALL", true]
+    ["Tracklist own/current show","AUTH_TRACKLIST_OWN", true],
+    ["Tracklist for any show/timeslot","AUTH_TRACKLIST_ALL", true]
 
 ]
--- a/schema/data-auth.json
+++ b/schema/data-auth.json
@@ -78,4 +78,7 @@
     ["Stop Broadcast","AUTH_STOPBROADCAST", true],
     ["Add Quotes","AUTH_ADD_QUOTES", true],
     ["Upload Music - Manual","AUTH_UPLOADMUSICMANUAL", true]
+    ["User can tracklist own/current show","AUTH_TRACKLIST_OWN", true],
+    ["User can tracklist for any show/timeslot","AUTH_TRACKLIST_ALL", true]
+
 ]
--- a/src/Classes/ServiceAPI/MyRadio_Timeslot.php
+++ b/src/Classes/ServiceAPI/MyRadio_Timeslot.php
@@ -258,8 +258,9 @@ class MyRadio_Timeslot extends MyRadio_Metadata_Common
             $timeslot = self::getInstance($_SESSION['timeslotid']);
 
             $result = [
-                "time" => $timeslot->getStartTime(),
-                "timeslotid" => $timeslot->getId()
+                "starttime" => $timeslot->getStartTime(),
+                "timeslotid" => $timeslot->getId(),
+                "title" => $timeslot->getMeta("title")
             ];
             return $result;
         }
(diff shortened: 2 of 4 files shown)


schema/data-auth.json
```diff
@@ -78,4 +78,7 @@
     ["Stop Broadcast","AUTH_STOPBROADCAST", true],
     ["Add Quotes","AUTH_ADD_QUOTES", true],
     ["Upload Music - Manual","AUTH_UPLOADMUSICMANUAL", true]
+    ["User can tracklist own/current show","AUTH_TRACKLIST_OWN", true],
+    ["User can tracklist for any show/timeslot","AUTH_TRACKLIST_ALL", true]
```
markspolakovs: Is it worth making SIS use these permissions as well? Otherwise it's a bit inconsistent

markspolakovs: Also is it just me, or is the permission naming inconsistent? None of the others have the `User can` prefix

src/Classes/ServiceAPI/MyRadio_Timeslot.php
```diff
@@ -248,6 +248,24 @@ public function getTimeslotAfter($filter = [1])
...
+    *
+    * @return array  Time, id If null, no timeslot is selected/user is logged out.
+    */
+    public static function getUserSelectedTimeslot() {
+        if (isset($_SESSION['timeslotid'])) {
+            $timeslot = self::getInstance($_SESSION['timeslotid']);
+
+            $result = [
+                "time" => $timeslot->getStartTime(),
+                "timeslotid" => $timeslot->getId()
```
markspolakovs: Worth maybe returning the complete data about the timeslot (title, credits etc.), instead of just the ID? Could save an extra API call in the future.

src/Classes/ServiceAPI/MyRadio_TracklistItem.php
```diff
@@ -70,6 +71,119 @@ protected static function factory($id)
...
+     *
+     * @throws MyRadioException
+     */
+    public static function create($trackid, $timeslotid = null, $starttime = null, $sourceid = 'a', $state = 'c')
+    {
+
+        if (AuthUtils::hasPermission(AUTH_TRACKLIST_ALL)) {
+            $tracklist_all = true;
+        }
+        else if (AuthUtils::hasPermission(AUTH_TRACKLIST_OWN))
```
markspolakovs: Braces

src/Classes/ServiceAPI/MyRadio_TracklistItem.php
```diff
@@ -70,6 +71,119 @@ protected static function factory($id)
...
+            $timeslot = MyRadio_Timeslot::getInstance($timeslotid);
+        }
+
+        if ($timeslot == null) {
+            // we're on jukebox
+            if ($tracklist_all == false) {
+                throw new MyRadioException("The current user doesn't have permission to set a tracklist on a show other than their own.", 403);
+            }
+        } else {
+            if ($tracklist_all == false && !in_array(MyRadio_User::getCurrentUser(), $timeslot->getSeason()->getShow()->getCreditObjects())) {
```
markspolakovs: Are `MyRadio_User` objects singletons? (that is, is it guaranteed that the object for user 123 in `getCreditObjects()` and the object that `getCurrentUser()` returns will be the same?

schema/data-auth.json
mstratford: Hmm, there's an overall need to consolidate all of the varying different functions that do the same all over the shop, but I want to say that's out of scope for this PR.

mstratford: Fair

src/Classes/ServiceAPI/MyRadio_Timeslot.php
mstratford: I don't really like the idea of returning the entire object, since this will be user-facing and could maybe return more info than expected. Hence the more limited return...

mstratford: I've added show title now though.

schema/data-auth.json
markspolakovs: Fair, file that as a future issue

src/Classes/ServiceAPI/MyRadio_Timeslot.php
markspolakovs: It's already user-facing though (e.g. https://ury.org.uk/api/v2/timeslot/146667)

mstratford: hmm

schema/data-auth.json
mstratford: #861

PR #862, opened 2020-04-02
Add WebStudio action and associated permissions

markspolakovs:
This PR adds an action to direct users to WebStudio.

I've already created an associated "WebStudio Trained" presenterstatus on prod, verified correct operation on Staging.

### DIFF
--- a/schema/data-actions.json
+++ b/schema/data-actions.json
@@ -68,6 +68,7 @@
     ["MyRadio", "listPermissions"],
     ["MyRadio", "permissionUsage"],
     ["MyRadio", "errorReport"],
+    ["MyRadio", "webstudio"],
     ["NIPSWeb", "default"],
     ["NIPSWeb", "create_token"],
     ["NIPSWeb", "secure_play"],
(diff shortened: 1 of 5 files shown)


mstratford:
Is it worth actually having a myradio controller for this? Webstudio already works out which timeslot you have selected...

markspolakovs:
Yes, just for the sake of "go to myradio and click webstudio" instead of "go to myradio, select a timeslot, then type in this other URL"

mstratford:
Oh yeah, I did mean to keep the button, but we don't need to put the timeslot_id in the URL anymore anyway. Could just link directly to /webstudio

markspolakovs:
Ah, fair. The benefit of the controller is that we can use `CoreUtils::requireTimeslot();`, saving a redirect ping-pong of MyRadio->WebStudio->MyRadio->WebStudio

mstratford:
Meh, fair. Maybe we just remove the timeslot_id from the URL, but still controller. Looks cleaner to user, and currently typing in a timeslot_id will make the show plan loaded not line up with the navbar data.

markspolakovs:
Sure, let's do that.

PR #864, opened 2020-04-05
Add some new user playlists

markspolakovs:


michael-grace:
still should have called it MyAlan or MyTony

mstratford:
Hmm, i'm probably not the best for "doing things not professionally"

PR #866, opened 2020-04-13
Fix tracklisting during DST.

mstratford:


markspolakovs:
Looks fine in theory, I presume you've tested it

PR #867, opened 2020-04-14
Correct timezone calculation to only affect the DB table.

mstratford:


markspolakovs:
Should be alright, have you tested it?

mstratford:
It’s what actually started tracklisting in production last night :)

markspolakovs:
Oh good.

PR #868, opened 2020-04-19
Disable sessions for CLI controller

markspolakovs:
Many of the "sso_session" duplicate errors we've been getting have a common denominator - they're called by root_cli.php. Thing is, CLI actions shouldn't need sessions, as they run as Mr Website... right?

### DIFF
--- a/src/Controllers/root_cli.php
+++ b/src/Controllers/root_cli.php
@@ -2,4 +2,7 @@
 /**
  * Sets up the environment for MyRadio non-web Controllers.
  */
+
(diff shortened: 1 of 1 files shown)


markspolakovs:
I'll be honest, this is just a hunch, I'm not actually sure if this'll work. We might need the old NullSession nonsense (you know, the one that we removed because it broke all of SIS for no reason...) here

markspolakovs:
Ugh, API uses `root_cli`, doesn't it. Bloody hell.

PR #870, opened 2020-04-21
Add include_pending to allpodcasts API, so website can ignore them.

mstratford:
MyRadio-go PR to match.


src/Classes/ServiceAPI/MyRadio_Podcast.php
```diff
@@ -855,13 +856,27 @@ public function convert()
...
+        if (!$include_suspended || !$include_pending) {
+            $where = 'WHERE ';
+            if (!$include_suspended) {
+                $where .= 'suspended = false';
+            }
+            if (!$include_suspended && !$include_pending) {
+                $where .= ' AND ';
+            }
+            if (!$include_pending) {
+                $where .= 'NOT submitted IS NULL';
```
markspolakovs, line 876: `submitted IS NOT NULL` reads a bit cleaner, IMHO

src/Classes/ServiceAPI/MyRadio_Podcast.php
```diff
@@ -855,13 +856,27 @@ public function convert()
...
      * @param int $num_results The number of results to return per page. 0 for all podcasts.
      * @param int $page The page required.
-     * @param bool $includeSuspended Whether to include suspended podcasts in the result
+     * @param bool $include_suspended Whether to include suspended podcasts in the result
+     * @param bool $include_pending Whether to include pending (future publish/processing) podcasts in the result
      *
      * @return Array[MyRadio_Podcast]
      */
-    public static function getAllPodcasts($num_results = 0, $page = 1, $includeSuspended = true)
+    public static function getAllPodcasts($num_results = 0, $page = 1, $include_suspended = true, $include_pending = true)
```
markspolakovs, line 864: Just to check that I'm understanding this right, `include_pending` defaults to TRUE because switching the default would be a breaking change (even though that behaviour is technically incorrect)?

If so, how many callsites are there that actually rely on pending being included? If there aren't that many, it make more sense to flip the default and correct them instead.

mstratford: Mmm, so, I believe the only call sites are the "all podcasts" page of myradio, and myradio-go/2016-site, so in theory there's not too much to change (a single controller in myradio and myradio-go). Could be done.

src/Classes/ServiceAPI/MyRadio_Podcast.php
mstratford: Mmm, yes, I think this was a testing leftover from a mistake of mine.

src/Classes/ServiceAPI/MyRadio_Podcast.php
mstratford: Done.

PR #874, opened 2020-04-25
CS restore training map

cs-499243:
Bear in mind this is my first time doing PHP so it can likely be adjusted. I've tested this out locally and it seems to display everything fairly okay. Query in the main PHP file gives the expected result in our database too.


cs-499243:
Actually I need to consider if the SQL query takes the right test value. This might show the names of people but not necessarily those with that training, depending on if I got it right. My test sample was 2 users with self-gained training so this could have caused that issue. All I would need to do is change the "WHERE" condition to the right value

mstratford:
I'm torn between essentially returning all the users as MyRadio_User so you can link to them, but I also know that there's a lot of overhead there for a list of people. So I'm not too fussed.

Issue #752, opened 2018-06-18
Related Show for Podcast list too long.


mstratford:
If you have the permissions to view all shows, selecting from all shows in the dropdown is tedious, especially since it's not in name order.

Brookke:
Could the drop down include a search box? https://dev.to/blacksonic/autocomplete-dropdown-using-native-html-moi

PR #876, opened 2020-04-25
If user can put a podcast to any show, limit to current term shows.

mstratford:
Fixes: #752


markspolakovs:
Mm, not sure this is such a good fix - I'd expect "any show" to really mean "any show". Also, this precludes the case of creating a show just for the sake of grouping podcasts (in a timeline where that's possible)

markspolakovs:
Closing now for reasons above and to clean up PRs. Am open to further discussion, as I do know the l0ng list of shows is an issue.

PR #877, opened 2020-04-26
Cs minor alterations 20 04 25

cs-499243:
Gives a password check to the setup and cleans up the display a little


src/Controllers/Setup/user.php
```diff
@@ -7,7 +7,7 @@
 use \MyRadio\MyRadio\CoreUtils;
 use \MyRadio\ServiceAPI\MyRadio_User;
 
-if ($_SERVER['REQUEST_METHOD'] === 'POST') {
+if ($_SERVER['REQUEST_METHOD'] === 'POST' && $_REQUEST['password'] === $_REQUEST['passwordchk']) {
```
markspolakovs, line 10: You probably want a descriptive error message

PR #878, opened 2020-04-27
email_pipe: use the proper spam status header

markspolakovs:
If a message is spam, Exim doesn't set the `X-Spam-Status: yes` header, but rather the `X-Spam_action: reject` header. Update the email_pipe controller to use the correct one.


markspolakovs:
@mstratford I'll be honest, I have no idea either - probably somewhere in the Exim config

PR #881, opened 2020-05-02
Reduce podcast bitrate to 128kbps

markspolakovs:
That's what most of the big players use (tested Serial and Relay FM), and frankly 192k is overkill for an hour of speech anyway


markspolakovs:
Merging. Is it worth reprocessing existing podcasts, or should we just leave the current 192k ones?

PR #883, opened 2020-05-06
Allow marking "podcast shows" as explicit

markspolakovs:


markspolakovs:
Tested creation and editing logic on dev, although for some reason the RSS API doesn't work on dev... hmm

mstratford:
Looks good, if we’re changing the podcasting policy to allow explicit stuff, we’ll have to make that obvious in the embed player or whatever in 2016-site too.

markspolakovs:
@mstratford it's always been allowed, policy just said "add a content warning to the description in bold". Could surface this in the UI, though.

markspolakovs:
Hm, this appears to have surfaced a bug in MyRadio_Show::create... investigating

markspolakovs:
Test's funky, need to figure out why. Works in staging, merging.

PR #884, opened 2020-05-07
Development Dockerfile

markspolakovs:


mstratford:
Is this ready to review? :)

markspolakovs:
In theory

markspolakovs:
Eh, good enough, we can always iterate

cs-499243:
```
Config::$short_name = 'URN';
Config::$long_name = 'University Radio Nork';
```
Very nice

Also is this going to be sure to use the right postgres version? Just asking cause bootstrap.sh is explicit about it

markspolakovs:
> Also is this going to be sure to use the right postgres version? Just asking cause bootstrap.sh is explicit about it

It’ll use whichever is tagged as latest on https://hub.docker.com/_/postgres, which right now is 14… so we might want to pin it to a Postgres major at least.

PR #888, opened 2020-05-17
Allow for multiple show photo sizes, and resize on upload

markspolakovs:
Note that this changes the config variable $default_show_url to $default_show_path, and will need to be updated in prod to match.

**Untested, do not merge.**

### DIFF
--- a/src/Classes/Config.php
+++ b/src/Classes/Config.php
@@ -231,6 +231,19 @@ final class Config
     public static $webcam_set_url;
 
     /**
+     * The formats to resize uploaded images to.
+     *
+     * The keys of the array are format names to rename to. The values are [x,y] sizes in pixels.
+     * @var int[][]
+     */
+    public static $image_resize_formats = [
+        'podcast_small' => [1400, 1400],
+        'web' => [800, 800]
+    ];
+
+    public static $default_image_format = 'web';
+
+    /**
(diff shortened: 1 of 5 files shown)


src/Classes/Config.php
```diff
@@ -253,12 +266,12 @@ final class Config
...
 
     /**
-     * The full web address to the image that will be served for a show if there
+     * The local path, relative to $public_media_path, to the image that will be served for a show if there
      * is not a photo for that show.
      *
      * @var string
      */
-    public static $default_show_uri = '/media/image_meta/ShowImageMetadata/22.png';
+    public static $default_show_path = 'image_meta/ShowImageMetadata/22.png';
```
mstratford, line 275: if we're doing this, mind changing this to have "image" or "photo" etc in the var name? :)

markspolakovs:
@mstratford good point on the jukebox one, have to check that. Re seasons, iirc seasons inherit from shows (which is kinda dumb but okay) so they should use the show's method, although I need to double-czech. Either way, I can't think of all that many places where we use season photos, usually it's either show or timeslot.

mstratford:
Oh yeah, Indeed, we had season pages in 2013-site, but they were just dumb, so we removed them in favour of the column thing on the show page.

PR #890, opened 2020-05-18
PHPUnit tests, at long last

markspolakovs:


markspolakovs:
hrrrng, phpunit 9 doesn't support php < 7.3. Worth a downgrade to phpunit 8?

PR #895, opened 2020-05-24
Semi-Anonymous page view analytics

markspolakovs:
This PR adds semi-anonymous page view tracking. Intention is to figure out which pages people actually use, and how they get there, so that we have data to go on for MyRadio^2.

Some notes:

- API is excluded, for obvious reasons.
- If you're not an officer, this is essentially anonymous. All the identifying info that is stored is that you're not an officer, and the number of shows you do, bucketed by five and rounded up.
- If you are an officer, this can de-anonymise you, as it stores the officerships you have. Committee will need to consent, other officers should be fine unless they have a _unique_ officership _right now_ (as in, yes, you're a computing officer, one of the 15 - technically de-anonymising, but not enough for me to get worried).
- Session ID does deanonymise you, but only until it expires and gets pruned. If we really wanted to be careful with this, we could hash it I suppose.
- This _pounds_ the database. I used a stored procedure to only have one extra query on each page load instead of three, but it's still an extra query on each load. Plus, it's an infinitely growing table (although it's not like we don't have any of those...)

Currently running on myradio-dev for testing.


src/Classes/MyRadio/MyRadioAnalytics.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace MyRadio\MyRadio;
+
+class MyRadioAnalytics {
+    public static function createAnalyticsRecord($page, $referrer) {
```
markspolakovs, line 6: Whoops. This can go.

PR #897, opened 2020-05-27
Ensure seasons are always in chronological order

markspolakovs:


markspolakovs:
Live on staging, appears to not explode

PR #898, opened 2020-06-15
added slack to menu

ColinRoitt:
add slack to menu


cs-499243:
As much as it doesn't change much should we use https://ury.slack.com/ as the link?

ColinRoitt:
> Christ, Colin, what have you done to the diff?

listen I don't make the rules about what vscode formatting extensions try to do... its a losing battle out there...

LordAro:
So configure it better.

PR #899, opened 2020-06-19
Use credits for the timeslot, not whole show.

mstratford:
Not tested yet, but should fix single timeslot credits from not being able to tracklist.


markspolakovs:
@mstratford Ping me when it's tested, I can review and merge then

mstratford:
Same technique is used elsewhere, bit hard to test. Seems correct tho.

PR #907, opened 2020-07-15
Add permission to let someone train anyone in anything

markspolakovs:
As part of the Great 2020 Training Reshuffle, we're adding the first training status that has multiple prerequisites. Now, I'm a lazy bastard who can't be bothered to write the code for that, so instead I'll just make that perm have no prerequisites, and instead give whoever can give it the ability to give anything ;)

This will need to be followed up with a SQL change:

```sql
INSERT INTO public.l_action (descr, phpconstant) VALUES ('Give Anyone Any Training Status', 'AUTH_AWARDANYTRAINING');
```


mstratford:
Ran the SQL.

PR #917, opened 2020-07-31
implement radio time

michael-grace:
duration for how long someone is _theoretically_ on radio, can be used in https://github.com/UniversityRadioYork/2016-site/issues/53

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_User.php
+++ b/src/Classes/ServiceAPI/MyRadio_User.php
@@ -675,21 +675,25 @@ class MyRadio_User extends ServiceAPI implements APICaller
     public function getRadioTime()
     {
         if (!$this->radioTime) {
-            $this->radioTime = self::$db->fetchColumn(
-                'SELECT sum(duration)         
-                FROM schedule.show_season_timeslot
-         INNER JOIN schedule.show_season USING (show_season_id)
-         INNER JOIN schedule.show_credit USING (show_id)
-         INNER JOIN sis2.member_signin USING(show_season_timeslot_id)
-         WHERE show_credit.creditid = $1
-         AND member_signin.memberid = $1
-         AND show_credit.effective_from <= show_season_timeslot.start_time
-         AND (show_credit.effective_to > show_season_timeslot.start_time OR show_credit.effective_to IS NULL)
-         AND show_credit.approvedid IS NOT NULL;',
-         [$this->getID()]
-            )[0];
+            try{
+                $this->radioTime = self::$db->fetchColumn(
+                    'SELECT sum(duration)         
+                    FROM schedule.show_season_timeslot
+            INNER JOIN schedule.show_season USING (show_season_id)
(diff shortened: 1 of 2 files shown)


URY-GitHub:
Thank you for this pull request, @michael-grace!

@colinroitt, would you mind reviewing this PR, please? Feel free to ask anyone for help if you're not quite sure what you're doing.

src/Classes/ServiceAPI/MyRadio_User.php
```diff
@@ -662,6 +669,32 @@ public function getOfficerships()
         return $this->officerships;
     }
 
+    /**
+     * Get the User's radio time, calculated from timeslot signins
+     */
+    public function getRadioTime()
+    {
+        if (!$this->radiotime) {
+            $this->radiotime = self::$db->fetchColumn(
```
markspolakovs, line 678: No error checking? Does fetchColumn yeet errors or just return false or something like that?

src/Classes/ServiceAPI/MyRadio_User.php
```diff
@@ -2157,6 +2190,7 @@ public function toDataSource($mixins = [])
         $data['photo'] = $this->getProfilePhoto() === null ?
             Config::$default_person_uri : $this->getProfilePhoto()->getURL();
         $data['bio'] = $this->getBio();
+        $data['radio-time'] = $this->getRadioTime();
```
markspolakovs, line 2193: This isn't a valid JS identifier, so accessing it from `JSON.parse()` is pain. Consider `radioTime` or `radio_time`.

src/Classes/ServiceAPI/MyRadio_User.php
```diff
@@ -172,6 +172,13 @@ class MyRadio_User extends ServiceAPI implements APICaller
      */
     private $shows;
 
+    /**
+     * Users radio time, calculated from signed in shows
+     * 
+     * @var mixed
+     */
+    private $radiotime;
```
markspolakovs, line 180: Camel-casing, please.

src/Classes/ServiceAPI/MyRadio_User.php
```diff
@@ -662,6 +669,32 @@ public function getOfficerships()
         return $this->officerships;
     }
 
+    /**
+     * Get the User's radio time, calculated from timeslot signins
+     */
+    public function getRadioTime()
+    {
+        if (!$this->radioTime) {
+            $this->radioTime = self::$db->fetchColumn(
```
markspolakovs, line 678: So... currently this will actually yeet an exception if the query fails. However, the code that does that is labeled `// TODO: temporary, uncomment this - marks.polakovs` - so normally it just returns an empty array, so this will explode - but with a cryptic error. Approving nonetheless, but consider fixing it :)

michael-grace: shoved another commit with the catch, just said its null, unless we've got something to rethrow it to

PR #919, opened 2020-08-02
Allow ordering and hiding mailing lists in the UI

markspolakovs:


markspolakovs:
Schema patch applied cleanly on apitest, code currently live at https://ury.org.uk/myradio-dev/Mail/.

PR #922, opened 2020-08-11
Add MyRadio_TracklistItem::getNowPlaying

markspolakovs:
Possible since the Jukebox change I rolled out earlier today.

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Track.php
+++ b/src/Classes/ServiceAPI/MyRadio_Track.php
@@ -1573,6 +1573,7 @@ class MyRadio_Track extends ServiceAPI
                 AND timestopped IS NULL
                 AND trackid IS NOT NULL
                 AND timeplayed <= NOW() AND timeplayed > (NOW() - interval \'30 minutes\')
+                ORDER BY timeplayed DESC
                 LIMIT 1
                 ',
                 [ $selAction ]
--- a/src/Classes/ServiceAPI/MyRadio_Track.php
+++ b/src/Classes/ServiceAPI/MyRadio_Track.php
@@ -1561,7 +1561,7 @@ class MyRadio_Track extends ServiceAPI
             LIMIT 1',
             []
         );
-        $selAction = $result[0] ?? 0;
+        $selAction = isset($result[0]) ? intval($result[0]) : 0;
         if ($selAction === 4 /* Studio 1 */ || $selAction === 5 /* Studio 2 */ || $selAction == 7 /* OB */) {
             // Ditto on the 30 minutes
             // The 30 *seconds* is to (hopefully) catch PFLs
--- a/src/Classes/ServiceAPI/MyRadio_Track.php
+++ b/src/Classes/ServiceAPI/MyRadio_Track.php
@@ -1564,6 +1564,7 @@ class MyRadio_Track extends ServiceAPI
         $selAction = $result[0] ?? 0;
         if ($selAction === 4 /* Studio 1 */ || $selAction === 5 /* Studio 2 */ || $selAction == 7 /* OB */) {
             // Ditto on the 30 minutes
+            // The 30 *seconds* is to (hopefully) catch PFLs
             $lastBapsLogged = self::$db->fetchOne(
                 'SELECT audiologid, timeplayed AS timestart, trackid
                 FROM public.baps_audiolog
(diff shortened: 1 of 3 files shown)


URY-GitHub:

@colinroitt, would you mind reviewing this PR, please? Feel free to ask anyone for help if you're not quite sure what you're doing.

markspolakovs:
To test, https://ury.org.uk/api-staging/graphql?query={nowPlaying(include_playout:true){startTime,track{...%20on%20Track{title, artist},...%20on%20TrackNotRec{title,artist}}}}

markspolakovs:
Hrm, if you set `include_playout` to `false` in the above query, it thinks Midnight City is on air - because it was the last thing that was played by Jukebox when on air. Do we want to backfill `timestop` data in the database, or will it break shit? (Or do we want to take the easy way out and do `AND timestart > '2020-08-11 17:00:00`?)

michael-grace:
i mean, omitting the api key still lets us get the info, do we want that?

markspolakovs:
@michael-grace I mean the API key is gonna be public in radioplayer anyway, so I'm not sure if it makes much difference - all the more sensitive endpoints are (read: should be!) gated behind perms

michael-grace:
also, i think backfilling is the best solution,

`update tracklist.tracklist
set timestop = tracklist.timestart +
               (select rec_track.length from public.rec_track where rec_track.trackid = (select track_rec.trackid from tracklist.track_rec where track_rec.audiologid = tracklist.audiologid))
where tracklist.timestop = null` should work, but might bug for track_notrec

markspolakovs:
@michael-grace couldn't you just do (can't be bothered to write out proper sql but) `update tracklist set timestop = (timestart of next one) where timestop is null`?

michael-grace:
no, because manual additions don't have timestop, and they aren't neccesary sequential, say with 'o' from jukebox. i mean, it probably doesn't really matter

markspolakovs:
Bleh, yeah, your approach works too. Let's do that, and just eyeball 3 minutes or such for manual ones

ColinRoitt:
Still want my review on this?

markspolakovs:
@ColinRoitt if you have the time why not - more pairs of eyes and all that

mstratford:
So, looking at https://github.com/UniversityRadioYork/icetracks/blob/master/icetracks.sh (which this is effectively replacing)...

I notice another BAPSism we’re currently ignoring with this replacement is the “BAPS cueing tracklists normally” 10 secs buffer. This is because BAPS can’t tell if you’re playing on PFL or not, so a 10 sec grace period is set.

I believe I’ve checked this before, but I presume the replacement of the baps tracklists are fine since it’s mapped into the new general table.

markspolakovs:
@mstratford I believe so, `bapstotracklist` should take care of it (see https://ury.org.uk/ceedox/computing:software:in-house:tracklisting)

src/Classes/ServiceAPI/MyRadio_Track.php
```diff
@@ -1527,6 +1527,91 @@ public static function getLibraryStats()
...
+                INNER JOIN tracklist.selbaps ON baps_audiolog.serverid = selbaps.bapsloc
+                WHERE selaction = $1
+                AND timestopped IS NULL
+                AND trackid IS NOT NULL
+                AND timeplayed <= NOW() AND timeplayed > (NOW() - interval \'30 minutes\')
+                ',
+                [ $selAction ]
+            );
+            if (!empty($lastBapsLogged)) {
+                if (strtotime($lastBapsLogged['timestart']) > strtotime($lastTracklisted['timestart'])) {
```
mstratford, line 1583: Do we need a check here in case the first query doesn't return something?

markspolakovs: Ah yes, good point

markspolakovs:
@SplashyFish14 #oops

Correct test link https://ury.org.uk/api-staging/graphql?query=%7BnowPlaying(include_playout:true)%7Bstart_time,track%7B...%20on%20Track%7Btitle,artist%7D,...%20on%20TrackNotRec%7Btitle,artist%7D%7D%7D%7D&api_key=API_KEY - appears to be sane, although I haven't tested it in the presence of a baps

markspolakovs:
So... the BAPS part still gets confused by DST. I get the feeling https://github.com/UniversityRadioYork/MyRadio/pull/922/files#diff-df7f98441b18206e232b40080c1880e6R1583is to blame.

mstratford:
So, we did have to fix one part of MyRadio for making Webstudio tracklists on mixclouder during DST, I'm still interested on if the fix will break when back on UTC.

markspolakovs:
Okay, verified, the BAPS module works.
image

mstratford (approved):

PR #924, opened 2020-08-15
Use is_in_byline to generate presenter string.

mstratford:
This allows co-presenters (and others, as specified in the credits db table) to appear in the presenters string used by radioplayer etc. Replaces #800.


URY-GitHub:

markspolakovs:
I will note that currently Presenter is the only credit_type with is_in_byline set... but this is a good change nonetheless

mstratford:
Indeed, currently trying to test it.

mstratford:
Ready 👍 Tested on Dev.

mstratford:
> I will note that currently Presenter is the only credit_type with is_in_byline set... but this is a good change nonetheless

I've just added Co-Presenters, which was the original issue.

PR #925, opened 2020-08-16
Prevent longer than 20char tags

mstratford:
A dormant branch that I just found the code for. Not sure if we need to put season / timeslot stuff in too.

Mainly an issue because it looks / feels sensible. Seen some abuse that just looks a bit meh in the past.

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Podcast.php
+++ b/src/Classes/ServiceAPI/MyRadio_Podcast.php
@@ -403,6 +403,12 @@ class MyRadio_Podcast extends MyRadio_Metadata_Common
         MyRadio_Show $show = null,
         $credits = null
     ) {
+        //Explode the tags
+        $tags = explode(' ', $tags);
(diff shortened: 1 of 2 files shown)


URY-GitHub:

mstratford:
Oh, hang on, this may have been re-done already.

src/Classes/ServiceAPI/MyRadio_Podcast.php
```diff
@@ -403,6 +403,12 @@ public static function create(
         MyRadio_Show $show = null,
         $credits = null
     ) {
+        //Explode the tags
+        $tags = explode(' ', $tags);
```
markspolakovs, line 407: Shouldn't this use CoreUtils::explodeTags?

mstratford: I think coreUtils::explodeTags was created after this, and supersedes this PR, closing 👍

PR #927, opened 2020-08-16
Implement GraphQL mutations

markspolakovs:
Probably about time I got this merged, so this branch doesn't have to live on for any longer

### DIFF
--- a/schema/api.graphql
+++ b/schema/api.graphql
@@ -241,6 +241,11 @@ type ShowCredit {
     User: User
 }
 
+type Genre {
+    value: Int!
+    text: String!
+}
+
 type Show implements Node & MyRadioObject @auth(hook: ViewShow) {
     id: ID!
     itemId: Int! @bind(method: "getID")
@@ -336,6 +341,8 @@ type Query {
 
     allShows(current_term_only: Boolean): [Show] @bind(class: "\\MyRadio\\ServiceAPI\\MyRadio_Show", method: "getAllShows")
 
+    allSubtypes: [ShowSubtype!] @bind(class: "\\MyRadio\\ServiceAPI\\MyRadio_ShowSubtype", method: "getAll") @auth(constants: [])
+
     currentTimeslot: Timeslot @bind(class: "\\MyRadio\\ServiceAPI\\MyRadio_Timeslot", method: "getCurrentTimeslot")
     nextTimeslot: Timeslot @bind(class: "\\MyRadio\\ServiceAPI\\MyRadio_Timeslot", method: "getNextTimeslot")
     nineDaySchedule(weekno: Int!, year: Int): Timeslot @bind(class: "\\MyRadio\\ServiceAPI\\MyRadio_Timeslot", method: "get9DaySchedule")
@@ -346,6 +353,8 @@ type Query {
 
     allMailingLists: [MailingList!] @bind(class: "\\MyRadio\\ServiceAPI\\MyRadio_List", method: "getAllLists")
     mailingListByName(name: String!): MailingList @bind(class: "\\MyRadio\\ServiceAPI\\MyRadio_List", method: "getByName")
+
+    allGenres: [Genre!] @bind(class: "\\MyRadio\\ServiceAPI\\MyRadio_Scheduler", method: "getGenres") @auth(constants: [])
 }
 
 type Mutation {
--- a/schema/api.graphql
+++ b/schema/api.graphql
@@ -358,7 +358,7 @@ input ShowCreditInput { # this is icky
 }
 
 input CreateShowInput {
-    name: String!
+    title: String!
     description: HTMLString!
     credits: ShowCreditInput!
     genres: [Int!]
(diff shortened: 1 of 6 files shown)


URY-GitHub:

@mstratford, would you mind reviewing this PR, please? Feel free to ask anyone for help if you're not quite sure what you're doing.

schema/api.graphql
```diff
@@ -452,8 +453,64 @@ type Query {
 
     allGenres: [Genre!] @bind(class: "\\MyRadio\\ServiceAPI\\MyRadio_Scheduler", method: "getGenres") @auth(constants: [])
     allCreditTypes: [CreditType!] @bind(class: "\\MyRadio\\ServiceAPI\\MyRadio_Scheduler", method: "getCreditTypes") @auth(constants: [])
+
+    findMemberByName(name: String!, limit: Int): [MemberSearchResult!] @bind(class: "\\MyRadio\\ServiceAPI\\MyRadio_User", method: "findByName") @auth(constants: ["AUTH_APPLYFORSHOW"]) # TODO not 100% sure this is the best constant
+    findShowByTitle(term: String!, limit: Int!): [FindShowByTitleResult!] @bind(class: "\\MyRadio\\ServiceAPI\\MyRadio_Scheduler", method: "findShowByTitle") @auth(constants: ["AUTH_APPLYFORSHOW"]) # TODO ditto
+}
+
+type Mutation {
+    createShow(input: CreateShowInput): Show @bind(class: "\\MyRadio\\ServiceAPI\\MyRadio_Show", method: "create", callingConvention: FirstArgInput)
```
markspolakovs, line 462: I'm not sure how I feel about the `FirstArgInput` hack - Relay's style guide recommends having all mutations take one argument called `input`, but I feel like getting that to work with the ol' PHP stack will require more fuckery than it's worth.

markspolakovs: On the other hand, compare

```gql
mutation MyMutation($input: UpdatePostInput!) {
  updatePost(input: $input) { ... }
}
```
and
```gql
mutation MyMutation($id: ID!, $newText: String, ...) {
  updatePost(id: $id, newText: $newText, ...) { ... }
}
```

markspolakovs: I'll probably end up adding another calling convention that "splats" the input type across the arguments, kinda like we already do with query args.

markspolakovs: Done with `InputAsArgs`.

schema/api.graphql
```diff
@@ -462,8 +464,70 @@ type Query {
 
     allGenres: [Genre!] @bind(class: "\\MyRadio\\ServiceAPI\\MyRadio_Scheduler", method: "getGenres") @auth(constants: [])
     allCreditTypes: [CreditType!] @bind(class: "\\MyRadio\\ServiceAPI\\MyRadio_Scheduler", method: "getCreditTypes") @auth(constants: [])
+
+    findMemberByName(name: String!, limit: Int): [MemberSearchResult!] @bind(class: "\\MyRadio\\ServiceAPI\\MyRadio_User", method: "findByName") @auth(constants: ["AUTH_APPLYFORSHOW"]) # TODO not 100% sure this is the best constant
+    findShowByTitle(term: String!, limit: Int!): [FindShowByTitleResult!] @bind(class: "\\MyRadio\\ServiceAPI\\MyRadio_Scheduler", method: "findShowByTitle") @auth(constants: ["AUTH_APPLYFORSHOW"]) # TODO ditto
+}
+
+#input ShowCreditInput { # this is icky
```
mstratford, line 472: Lots of commented out stuff :/

schema/api.graphql
```diff
@@ -347,7 +349,7 @@ type Timeslot implements Node & MyRadioObject @auth(hook: ViewShow) {
     photo: String
     messages: [Message]
     webpage: String!
-
+    uploadState: String @meta(key: "upload_state")
```
mstratford, line 352: duplicated?

schema/api.graphql
```diff
@@ -462,8 +464,70 @@ type Query {
 
     allGenres: [Genre!] @bind(class: "\\MyRadio\\ServiceAPI\\MyRadio_Scheduler", method: "getGenres") @auth(constants: [])
     allCreditTypes: [CreditType!] @bind(class: "\\MyRadio\\ServiceAPI\\MyRadio_Scheduler", method: "getCreditTypes") @auth(constants: [])
+
+    findMemberByName(name: String!, limit: Int): [MemberSearchResult!] @bind(class: "\\MyRadio\\ServiceAPI\\MyRadio_User", method: "findByName") @auth(constants: ["AUTH_APPLYFORSHOW"]) # TODO not 100% sure this is the best constant
```
mstratford, line 468: I guess out of scope, but the fact this uses a completely separate auth list to V2 makes me a bit sad.

mstratford (commented):
Not 100% how all this works, but have some comments.

schema/api.graphql
markspolakovs: It still defaults to V2 auth unless you override it.

The reason for this is that V2 auth is evaluated once (when you call the endpoint), while GraphQL auth is evaluated for every sub-method in the graph. The reason for *that* is that some object properties are really just methods (for example `User.phoneNumber`), which need an auth check of their own (for example, you may want anyone to be able to see basic details of a user, but only see the phone number if you or they are on committee). In V2 this is handled either by changing the shape of the response (in the former case) or by mixins (in the latter) - neither of which are a viable solution for GraphQL because of the strict typing.

schema/api.graphql
markspolakovs: woops

PR #928, opened 2020-08-23
User profile end-of-life

markspolakovs:


src/Classes/ServiceAPI/MyRadio_User.php
```diff
@@ -2052,6 +2078,93 @@ public function grantPermission($authid, $from = null, $to = null)
         }
     }
 
+    /**
+     * Deactivates this user (EOL Tier 1). They will not be able to sign in, receive email, receive credits for
+     * any shows, and any officerships will be ended.
+     *
+     * Note: DO NOT CALL THIS DIRECTLY outside of the relevant daemon! Instead use requestDeactivation, as it
```
markspolakovs, line 2085: Coming Soon:tm:

src/Classes/ServiceAPI/MyRadio_User.php
```diff
@@ -2052,6 +2078,93 @@ public function grantPermission($authid, $from = null, $to = null)
...
+        self::$cache->purge();
+
+        // And send a goodbye email
+        // We use mail() manually to avoid hacking with the innards of MyRadioEmail
+        // (as it'd refuse to send due to receive_email being false)
+        // TODO: is this a good idea?
+        $domain = Config::$email_domain;
+        $long_name = Config::$long_name;
+        $short_name = Config::$short_name;
+        mail(
```
LordAro, line 2142: There must be an existing send mail function somewhere

markspolakovs: The comment immediately above:

> We use mail() manually to avoid hacking with the innards of MyRadioEmail (as it'd refuse to send due to receive_email being false)

You do have a valid point though, and it may be better to refactor MRE to allow "send immediately, even if they've opted out".

cs-499243 (commented):
Looks good - can't say I can confirm the php entirely (so I won't approve) but I don't see any obvious issues throughout

PR #931, opened 2020-09-20
Limit the number of search results returned

markspolakovs:
Currently, https://ury.org.uk/api/v2/show/searchmeta/a is a one-way ticket to DoS city.

I would like to switch this to use Postgres' built-in FTS, but the metadata system makes this a bit tricky.

h/t @cs-499243 for pointing this out

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ src/vendor
 composer.lock
 node_modules/
 src/Public/img/stats_training_*.svg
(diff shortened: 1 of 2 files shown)


markspolakovs:
Test: https://ury.org.uk/api-staging/v2/show/searchmeta/music

michael-grace (approved):

PR #933, opened 2020-09-21
Add event editing form

markspolakovs:
This will require the creation of two permissions, `AUTH_CREATEEVENT` and `AUTH_EDITANYEVENT`. [edit: done on prod/staging]


URY-GitHub:

src/Classes/ServiceAPI/MyRadio_Event.php
```diff
@@ -252,4 +380,17 @@ public function toDataSource($mixins = [])
             'master' => (is_null($this->masterId) ? null : self::getInstance($this->masterId)->toDataSource($mixins))
         ];
     }
+
+    public function canWeEdit()
+    {
+        return $this->getHost()->getID() !== MyRadio_User::getCurrentOrSystemUser()->getID()
```
michael-grace, line 386: Surely this should be EQUALS not NOT EQUAL as if host==current, they can edit

src/Controllers/Events/editEvent.php
```diff
@@ -0,0 +1,66 @@
...
+    } else {
+        // edit
+        /** @var MyRadio_Event $event */
+        $event = MyRadio_Event::getInstance($data['id']);
+
+        // check permissions
+        $event->checkEditPermissions();
+
+        if ($data['title'] !== $event->getTitle()) {
+            $event->setTitle($data['title']);
```
michael-grace, line 30: Couldn't we just have an "update all", even if it passes some same-as-before info, to take this down to just 1 db hit

markspolakovs:
'Kay, I think all editing buttons work.

michael-grace (changes requested):
Left some commenteoes 🙂

src/Controllers/Events/editEvent.php
markspolakovs: Good point, will fix

src/Classes/ServiceAPI/MyRadio_Event.php
markspolakovs: oops

michael-grace (approved):

PR #938, opened 2020-09-25
Sign In Details

michael-grace:
Not the best way of implementing, but definitely the simplest. Not on any menus, cause in true MyRadio fashion, We ❤️️ Hidden Links (/Scheduler/tracking)

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Timeslot.php
+++ b/src/Classes/ServiceAPI/MyRadio_Timeslot.php
@@ -1141,7 +1141,7 @@ class MyRadio_Timeslot extends MyRadio_Metadata_Common
             array_map(
                 function ($x) {
                     return [
-                        'signedby' => $x['signerid'] ? MyRadio_User::getIn  ance($x['signerid']) : null,
+                        'signedby' => $x['signerid'] ? MyRadio_User::getInstance($x['signerid']) : null,
                         'location' => $x['location_name'],
                         'guest_info' => $x['guest_info'],
                         'time' => strtotime($x['sign_time'])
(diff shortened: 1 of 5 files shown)


URY-GitHub:
Thank you for this pull request, @michael-grace!

michael-grace:
could possibly be done with a selection thing to select what locations you're interested in viewing

markspolakovs:
@michael-grace that was exactly what I was thinking.

Also, consider filtering by time, instead of pulling up every single timeslot in the term? It may work initially, but with more studio shows it may become unworkable.

markspolakovs (approved):
Looks good for a MVP, although see comment above

markspolakovs:
Also, maybe add it to menu? It won't show up without appropriate perms anyway, so no point in hiding it.

markspolakovs:
image

You  may need to make `guest_info` HTML and add `<br>`s, as the form says "one on each line". (If you do that, be careful about XSS escaping.)

michael-grace:
image

src/Classes/ServiceAPI/MyRadio_Timeslot.php
```diff
@@ -1261,7 +1261,7 @@ public function signInGuests(string $guestInfo, $locationid)
         self::$db->query(
             'INSERT INTO sis2.guest_signin (show_season_timeslot_id, signerid, location, guest_info)
                 VALUES ($1, $2, $3, $4)',
-            [$this->getID(), MyRadio_User::getInstance()->getID(), $locationid, $guestInfo]
+            [$this->getID(), MyRadio_User::getInstance()->getID(), $locationid, htmlspecialchars($guestInfo, ENT_QUOTES)]
```
markspolakovs, line 1264: Screm - I'd rather put in the cleanest data possible, and format it when you display it. (Also, Twig might take care of escaping itself?)

markspolakovs: Wait, you use `html`, nvm - either way, move `htmlspecialchars` to Scheduler/tracking.php

michael-grace: nah, this seems to be the only way which works, remember say on quotes (rip) and show descriptions you can xss

PR #945, opened 2020-09-29
Add AUTH_USENIPSWEB check to MyrTimeslot::getShowPlan

markspolakovs:
To ensure that non-WS-Trained people can still access it.

**Note:** I reused AUTH_USENIPSWEB to avoid creating _yet another_ permission - so, when this is merged, we'll need to change auth for MyRadio_Timeslot::getShowPlan to NULL, and also reduce the number of people with AUTH_USENIPSWEB. In fact, now that I write this, might it be better to just add a new one, or rename AUTH_USENIPSWEB?


URY-GitHub:

cs-499243 (approved):

mstratford:
I'd rather make this use the "View any show" permission?

markspolakovs:
@mstratford good point, will change (and then we can deprecate USENIPSWEB)

PR #950, opened 2020-10-06
expose nipsweb item

michael-grace:
marks is going to complain that i haven't graphqlanised it


markspolakovs:
y

michael-grace:
z

michael-grace:
in seriousness, regarding making current "bad autoplayout" and probable "future better one" get managed item usefullness (i mean, atm the path)

markspolakovs:
Fair enough. Does this also account for managed playlists?

markspolakovs (approved):
Hm, interesting that review-machine didn't run. Oh well.

PR #951, opened 2020-10-06
Don't trust API to return 3 channel if they're empty.

mstratford:
Importer wouldn't import properly if the show plan didn't have at least something in all three channels.


URY-GitHub:

markspolakovs (approved):
I have no idea what this does or how this works, but it kinda looks sane?

SplashyFish14 (commented):
I have no idea what this does or how this works, but it kinda looks sane?

mstratford:
Essentially I was just adding the 3 weights from the three channels of the show plan. But you don't get all 3 show plan channels if they're not filled with anything. So the importer just went "i don't have a weight for this channel" and didn't import.

PR #954, opened 2020-10-08
Move "filter by track" above artist

cs-499243:
res-filter-artist-container makes me suspicious this might break the style of something. I wouldn't imagine it would take more than this to fix the order though.


URY-GitHub:
Thank you for this pull request, @cs-499243!

mstratford (approved):

PR #955, opened 2020-10-08
Some attempts at CI-related fixes

LordAro:
A (mildly) nice thing to spend my lunchbreak on. Probably needs further effort (the 500 from the test failure but no other error is weird)


URY-GitHub:
Thank you for this pull request, @LordAro!

LordAro:
Finally. (Formatting my own)
```
 {
   "status":"FAIL",
   "payload":"Query failure: INSERT INTO schedule.show (show_type_id, submitted, memberid, podcast_explicit)
                                            VALUES ($1, NOW(), $2, $3::boolean) RETURNING show_id
              Params: array (0 => 1, 1 => 1, 2 => 0)
              ERROR:  column \"podcast_explicit\" of relation \"show\" does not exist
              LINE 1: ...schedule.show (show_type_id, submitted, memberid, podcast_ex..."
}
```

.travis.yml
```diff
@@ -22,9 +22,11 @@ addons:
     postgresql: '9.5'
 services:
     - postgresql
+before_install:
+  - nvm install 12.19  # Latest LTS, for eslint
```
markspolakovs, line 26:
```suggestion
  - nvm install --lts  # Latest LTS, for eslint
```

markspolakovs (approved):
LGTM

.travis.yml
```diff
@@ -22,9 +22,11 @@ addons:
     postgresql: '9.5'
```
markspolakovs, line 22:
```suggestion
    postgresql: '12'
```

Might as well

LordAro: Well ideally you should be testing with the oldest version supported. Is there any particular need to bump it further?

mstratford (approved):
LGTM 👍

.travis.yml
mstratford: I'm not too fussed, if we keep it at the lowest, it'll be obvious when new features aren't compatible with the lowest version. Seems sensible.

markspolakovs: I did run into some incompatibilities when doing the 10->12 upgrade wrt. `WITH OIDS`, but not sure if that was a quirk of `pg_dump`'s output (read: I forgot some important flag somewhere), or actual breaking changes that'd bork `bsae.sql`.

markspolakovs: Ah yes, `base.sql` uses `default_with_oids`, which is deprecated since 11, and explicit `WITH OIDS` is gone in 12. I don't think we use them anywhere though.

LordAro:
It's worth pointing out that I don't think applying the sql patches in CI should be necessary. base.sql should reflect the latest database schema and be "up to date", the patches should be provided to bring an older database up to date. That's probably a separate issue though.

LordAro:
image
Also tick these buttons. (Settings -> Branches -> master)

PR #957, opened 2020-10-12
Can now select which camera to watch in focus mode

ColinRoitt:
Previously this page appeared incomplete. Now allows user to select which view they wish to focus on.

### DIFF
--- a/src/Templates/Webcam/focus.twig
+++ b/src/Templates/Webcam/focus.twig
@@ -1,7 +1,7 @@
 {% extends 'stripe.twig' %}
+<link rel="stylesheet" type="text/css" href="{{baseurl}}css/myradio.webcamfocus.css">
 
 {% block stripecontent %}
-<link rel="stylesheet" type="text/css" href="{{baseurl}}css/myradio.webcamfocus.css">
 {% if live %}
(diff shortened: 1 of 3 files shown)


URY-GitHub:
Thank you for this pull request, @ColinRoitt!

src/Templates/Webcam/focus.twig
```diff
@@ -1,15 +1,17 @@
 {% extends 'stripe.twig' %}
 
 {% block stripecontent %}
+<link rel="stylesheet" type="text/css" href="{{baseurl}}css/myradio.webcamfocus.css">
```
markspolakovs, line 4: This won't work, because `stripecontent` is within `<body>`. You'll need to move this to `{% block head %}` (I think could be wrong).

markspolakovs (changes requested):

src/Templates/Webcam/focus.twig
ColinRoitt: Tested locally and it works bcs HTML👏 but you're right does need to go above.

ColinRoitt:
Correction I've done it in now....

michael-grace (approved):
LGTM assuming it actually works

michael-grace:
apologies, they

PR #958, opened 2020-10-16
Allow presenters to select if their show should be played out

michael-grace:
Uses whatever is in channel 1. I know showplanner is slowly dying, but this can be chucked in on webstudio too. **don't merge until the system is actually up and ready to go**


src/Classes/ServiceAPI/MyRadio_Timeslot.php
```diff
@@ -50,7 +51,7 @@ protected function __construct($timeslot_id)
         // Note that credits have different metadata timeranges to text
         // This is annoying, but needs to be this way.
         $result = self::$db->fetchOne(
-            'SELECT show_season_timeslot_id, show_season_id, start_time, duration, memberid, (
+            'SELECT show_season_timeslot_id, show_season_id, start_time, duration, memberid, playout::boolean::text, (
```
markspolakovs, line 54: Why the double-cast?

markspolakovs (approved):
Not a huge fan of the whole system being called "playout" instead of something more specific but I guess it's a little late for that.

src/Classes/ServiceAPI/MyRadio_Timeslot.php
michael-grace: can't remember. i'd guess its cause it didn't work for some strange reason

PR #960, opened 2020-10-23
Php setup fix and stopping members pages being (as) slow

cs-499243:


src/Classes/ServiceAPI/MyRadio_TrainingStatus.php
```diff
@@ -312,9 +312,9 @@ public static function getAllAwardableTo(MyRadio_User $to, MyRadio_User $by = nu
 
         $statuses = [];
         foreach (self::getAll() as $status) {
-            if ((!$status->isAwardedTo($to))
+            if ($status->canAward($by)
                 && $status->hasDependency($to)
-                && $status->canAward($by)
+                && (!$status->isAwardedTo($to))
```
markspolakovs, line 317: Explain?

cs-499243: For some reason if you run canAward without the other tests it runs smoothly. Obviously I can't just remove the tests so I just put that one first so when that fails it doesn't test the other ones.

cs-499243: Not sure quite how significant it is but compare https://ury.org.uk/myradio/Profile/view/?memberid=11207 to https://ury.org.uk/myradio-dev/Profile/view/?memberid=11207 (ignoring the unrelated errors)

PR #971, opened 2021-01-06
Make SIS use GeoIP 2

markspolakovs:
Untested. Lol.

### DIFF
--- a/src/Classes/SIS/SIS_Utils.php
+++ b/src/Classes/SIS/SIS_Utils.php
@@ -7,8 +7,10 @@
 
 namespace MyRadio\SIS;
 
+use GeoIp2\Exception\AddressNotFoundException;
 use MyRadio\Config;
 use MyRadio\MyRadio\AuthUtils;
+use MyRadio\MyRadioGeoIP;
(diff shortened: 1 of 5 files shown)


src/Classes/SIS/SIS_Utils.php
```diff
@@ -61,13 +63,16 @@ public static function ipLookup($ip)
         $location = [];
 
         if (($query === null) or (pg_num_rows($query) == 0)) {
-            $geoip = geoip_record_by_name($ip);
-            if ($geoip === false) {
+            MyRadioGeoIP::wakeup();
+            try {
+                $geoip = MyRadioGeoIP::getInstance()->city($ip);
+                if (empty($geoip->city)) {
```
markspolakovs, line 69: This might need to be an `isset`? Not sure

markspolakovs:
Well... you can open SIS without spewing errors left and right. Don't know if IP resolution actually works though.

mstratford (commented):
Looks ok for someone who hasn't actually tested it :)

markspolakovs:
I'll merge it now just to avoid SIS being completely broken, then fix it if need be when the time comes.

PR #972, opened 2021-01-12
Fix deprecation error.

mstratford:


markspolakovs (approved):
Seems sane, but I'm questioning the need for this - is this a py2/3 thing?

LordAro:
Python 3.8 thing. See separation notice at the top of https://docs.python.org/3/library/collections.html

markspolakovs:
@LordAro I mean the reason for the `isinstance(reqpost.json, Callable)` check - surely if we're using a consistent version of requests, `Response.json()` would always have the same type?

LordAro:
Ah, yes

requests release notes:
>### 1.0.0 (2012-12-17)
>...
>Make Response.json() callable, not property.
>...

I think it can go, yes.

PR #973, opened 2021-01-16
Make setTracks work via the API

markspolakovs:
We can't pass in MyRadio_Track objects via the API, so let it accept IDs.

### DIFF
--- a/src/Classes/iTones/iTones_Playlist.php
+++ b/src/Classes/iTones/iTones_Playlist.php
@@ -370,16 +370,13 @@ class iTones_Playlist extends \MyRadio\ServiceAPI\ServiceAPI
      * @param string                $lockstr String that provides Write access to this Playlist.
  *                                              Acquired from acquireLock();
      * @param string|null           $notes   Optional. A textual commit message about the change.
-     * @param MyRadio_User|null     $user   User to attribute the update to
      *
      * @todo Push these changes to the playlist files on playoutsvc.ury.york.ac.uk. This should probably be a
      *       MyRadioDaemon configured to run only on that server.
      */
-    public function setTracks($tracks, $lockstr, $notes = null, MyRadio_User $user = null)
+    public function setTracks($tracks, $lockstr, $notes = null)
     {
-        if ($user === null) {
-            $user = MyRadio_User::getInstance();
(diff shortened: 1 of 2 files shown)


src/Classes/iTones/iTones_Playlist.php
```diff
@@ -366,9 +366,10 @@ public function validateLock($lockstr)
      *
      * Once that's done, go over every Track still in the temporary list and remove them from the Playlist
      *
-     * @param MyRadio_Track[] $tracks  Tracks to put in the playlist.
-     * @param string          $lockstr String that provides Write access to this Playlist. Acquired from acquireLock();
-     * @param string          $notes   Optional. A textual commit message about the change.
+     * @param MyRadio_Track[]|int[] $tracks  Tracks to put in the playlist.
+     * @param string                $lockstr String that provides Write access to this Playlist. Acquired from acquireLock();
+     * @param string|null           $notes   Optional. A textual commit message about the change.
+     * @param MyRadio_User|null     $user   User to attribute the update to
```
mstratford, line 372: How is the MyRadio_User decided here for the API? We should change it to use whoever the logged in user is, instead.

mstratford: Hum, I see it actually already does this, I presume the API won't let you feed in a user either, but why can you send a user in anyway?

markspolakovs: Look a few lines down, it already does that :) Maybe we should just get rid of `$user` though, as it doesn't  appear to be used anywhere and could pose some issues were this exposed to the API. However, https://github.com/UniversityRadioYork/MyRadio/blob/master/src/Classes/Daemons/MyRadio_PlaylistsDaemon.php#L76-L80 uses it...

markspolakovs: Fairly sure we can trivially fix this by changing this to use `getCurrentOrSystemUser()` and removing the `$user` param.

mstratford: Yeah, that's what I was thinking 👍

mstratford (approved):
LGTM

Issue #750, opened 2018-04-30
You can cancel timeslots in the past


LordAro:
You can retroactively cancel timeslots. Probably not a good idea to allow this

mstratford:
@michael-grace current PC thonks?

michael-grace:
Its useful for scheduling, but maybe not a good idea for your average presenter

PR #979, opened 2021-02-03
Better Scheduler

michael-grace:
This is me finally getting round to slowly fixing scheduler. Stuff will happen when I get round to it. Maybe before I finish being PC.


michael-grace:
1. Normal people can't cancel shows in past. Only programming admin can. Fixes #750

michael-grace:
2. Centralised, easy to use place for viewing and cancelling shows. Much more to add to this page

PR #980, opened 2021-02-08
The Grand Yeet

michael-grace:
RIP ShowPlanner

### DIFF
--- a/src/Templates/MyRadio/menu.twig
+++ b/src/Templates/MyRadio/menu.twig
@@ -33,6 +33,8 @@
   </div>
 {% endif %}
 
+<p>ShowPlanner has now been deprecated. Please use WebStudio to plan and rehearse your shows.</p> 
+
 {% for column in menu %}
 <div class="col-md-4">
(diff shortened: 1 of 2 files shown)


markspolakovs:
Should be deployed at the same time as https://github.com/UniversityRadioYork/WebStudio/pull/206 (and WS v1.5.2).

src/Templates/MyRadio/menu.twig
```diff
@@ -33,6 +33,8 @@
   </div>
 {% endif %}
 
+<p>ShowPlanner has now been deprecated. Please use WebStudio to plan and rehearse your shows.</p>
```
mstratford, line 36: Show Planner? With a space.

mstratford: Could we put this as a div with class="alert alert-info" plz

mstratford (approved):
A review, but I do have some nice to haves.

mstratford (approved):

PR #981, opened 2021-02-08
Imagine being able to edit news feeds, not just copying the old one ;)

michael-grace:
feeds latest news feed into new news feed. make updating PIS easy


ColinRoitt (approved):
Looks good, fine if tested

Issue #903, opened 2020-06-24
All menu pages opening in a new tab is annoying


markspolakovs:
Consider: only external links in a new tab?

Brookke:
Might be worth prefixing external links with some sort of icon e.g. https://fontawesome.com/v4.7.0/icon/external-link

michael-grace:
closed by #984

PR #984, opened 2021-02-23
Only new tabs for external links

SplashyFish14:
Makes it so only clicking on non MyRadio links will open in new tabs, any internal link will use the same tab


michael-grace:
Alex Towells: "please go to bed Alice"

michael-grace (approved):
Well it works.

Fixes #903

PR #990, opened 2021-02-26
Add Please Select location

SplashyFish14:
this only half fixes Michaels issue, sorry Michael


SplashyFish14:
The other half of the issue is to make it so you can't go to the next page without selecting an option but I didn't know how to do that

michael-grace:
What you'll want is src/Controllers/MyRadio/timeslot.php and in the section labelled submitted an if statement for if the location from $_POST is "Please select". Have a look at the if statement for not selecting a timeslot with the URLUtils back with message

michael-grace (approved):

PR #992, opened 2021-03-10
Fix GraphQL Types

michael-grace:
POLAKOVS!


markspolakovs (approved):
Whoops.

(in my defence I’ve probably fixed that in an unmerged branch somewhere…)

PR #994, opened 2021-03-13
Slow Down for Rate Limiting

michael-grace:
Slow explicit daemon down so we can rerun big batches without being rate limited by itunes


markspolakovs (approved):
Lol I didn't know this still worked

PR #998, opened 2021-04-05
Summer Term Track and Trace

michael-grace:
I still need to test the daemon, among other things

### DIFF
--- a/src/Classes/Daemons/MyRadio_TrackAndTraceDaemon.php
+++ b/src/Classes/Daemons/MyRadio_TrackAndTraceDaemon.php
@@ -23,10 +23,12 @@ class MyRadio_TrackAndTraceDaemon extends \MyRadio\MyRadio\MyRadio_Daemon
             return;
         }
 
-        self::generateTrackAndTraceReport();
-
-        // Done
-        self::setVal($weekkey, time());
+        try {
+            self::generateTrackAndTraceReport();
+        } finally {
+            // Done
+            self::setVal($weekkey, time());
+        }
     }
 
     private static function generateTrackAndTraceReport()
--- a/src/Classes/Daemons/MyRadio_TrackAndTraceDaemon.php
+++ b/src/Classes/Daemons/MyRadio_TrackAndTraceDaemon.php
@@ -1,0 +1,91 @@
+<?php
+
+namespace MyRadio\Daemons;
+
+use MyRadio\Config;
+use MyRadio\ServiceAPI\MyRadio_Timeslot;
+use MyRadio\ServiceAPI\MyRadio_Season;
+use MyRadio\MyRadio\CoreUtils;
+use MyRadio\MyRadioEmail;
+use MyRadio\ServiceAPI\MyRadio_List;
+
+class MyRadio_TrackAndTraceDaemon extends \MyRadio\MyRadio\MyRadio_Daemon
+{
+    public static function isEnabled()
+    {
+        return Config::$d_TrackAndTrace_enabled;
+    }
+
+    public static function run()
+    {
+        $weekkey = __CLASS__ . '_last_run_weekly';
+        if (self::getVal($weekkey) > time() - 604700) {
+            return;
+        }
+
+        self::generateTrackAndTraceReport();
+
+        // Done
+        self::setVal($weekkey, time());
+    }
+
+    private static function generateTrackAndTraceReport()
+    {
+        $table = "<table>";
+        $table .= "<tr><th></th><th>Information</th><th>Location</th><th>Time</th></tr>";
+
+        $data = [];
+        $no_track = MyRadio_Timeslot::getLocationName(5); //WebStudio
+
+        foreach (MyRadio_Season::getAllSeasonsInLatestTerm() as $season) {
+            foreach ($season->getAllTimeslots() as $timeslot) {
+                if (
+                    $timeslot->getStartTime() < time()
+                    && $timeslot->getStartTime() > time() - 604800
+                ) {
+                    foreach ($timeslot->getSigninInfo() as $info) {
+                        if (isset($info["location"]) && $info["location"] != $no_track) {
+                            if (isset($info["user"])) {
+                                $eduroam = $info["user"]->getEduroam();
+                                $data[] = [
+                                    "type" => "URY Member",
+                                    "info" => $info["user"]->getName() . ($eduroam ? " ($eduroam)" : ""),
+                                    "location" => $info["location"],
+                                    "time" => CoreUtils::happyTime($info["time"]),
+                                    "unix" => $info["time"]
+                                ];
+                            } elseif ($info["guest_info"]) {
+                                $data[] = [
+                                    "type" => "Guest",
+                                    "info" => [
+                                        "display" => "html",
+                                        "html" => nl2br($info["guest_info"])
+                                    ],
+                                    "location" => $info["location"],
+                                    "time" => CoreUtils::happyTime($info["time"]),
+                                    "unix" => $info["time"]
+                                ];
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
+        foreach ($data as $row) {
+            $table .= "<tr><td>" . $row["type"]
+                . "</td><td>" . $row["info"]
+                . "</td><td>" . $row["location"]
+                . "</td><td>" . $row["time"]
+                . "</td></tr>\r\n";
+        }
+
+        $table .= "</table>";
+
+        MyRadioEmail::sendEmailToList(
+            MyRadio_List::getByName("Management Team"),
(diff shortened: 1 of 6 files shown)


src/Classes/ServiceAPI/MyRadio_Season.php
```diff
@@ -371,6 +383,24 @@ public static function getForm()
...
+                'studio-request',
+                MyRadioFormField::TYPE_CHECK,
+                [
+                    'explanation' => 'If ticked, you\'ll be requesting to present your show from our studios.
+                    It is very important that you have read the 
+                    <a href="https://drive.google.com/file/d/1Th_ih_XibBNTDg9izxgYFIOenFBKqwDq/view?usp=sharing"
+                    target="_blank">Studio Usage Rules</a>, 
+                    including that you can only do a solo show, or have people from the same household as you.
+                    Also, you can only do a studio show if you\'ve been previously studio trained,
+                    as we can\'t do training.There is <b>no</b>
```
markspolakovs, line 397:
```suggestion
                    as we can\'t do training. There is <b>no</b>
```

src/Controllers/Scheduler/tracking.php
```diff
@@ -32,7 +33,8 @@
                                 "html" => nl2br($info["guest_info"])
                             ],
                             "location" => $info["location"],
-                            "time" => CoreUtils::happyTime($info["time"])
+                            "time" => CoreUtils::happyTime($info["time"]),
+                            "unix" => $info["time"]
```
markspolakovs, line 37: What's unix actually doing? I don't see a user for it.

michael-grace: well, sorting is alphabetical, which happyTime isn't suited for, so although the unix time column is hidden, its sorted by that, so its in time order, to make it less nightmare to deal with

markspolakovs: ah, you finally fixed that annoyance. Nice.

michael-grace:
Tested, but don't merge until we see from YUSU if we're gonna need anything else (probably next week)

src/Classes/Daemons/MyRadio_TrackAndTraceDaemon.php
```diff
@@ -71,6 +71,13 @@ private static function generateTrackAndTraceReport()
             }
         }
 
+        usort($data, function ($a, $b) {
+            if ($a["unix"] == $b["unix"]) {
+                return 0;
+            }
+            return ($a["unix"] < $b["unix"]) ? -1 : 1;
```
markspolakovs, line 78: Can (in theory) just be
```suggestion
            return $a["unix"] - $b["unix"];
```

(the proof is left as an exercise to the reader :P )

src/Classes/Daemons/MyRadio_TrackAndTraceDaemon.php
```diff
@@ -0,0 +1,97 @@
...
+    {
+        $weekkey = __CLASS__ . '_last_run_weekly';
+        if (self::getVal($weekkey) > time() - 604700) {
+            return;
+        }
+
+        self::generateTrackAndTraceReport();
+
+        // Done
+        self::setVal($weekkey, time());
```
markspolakovs, line 29: If generateTrackAndTraceReport yeets an exception, last_run won't be updated and myradio_daemon will loop infinitely, won't it? Is this intended behaviour?

src/Classes/Daemons/MyRadio_TrackAndTraceDaemon.php
```diff
@@ -0,0 +1,97 @@
...
+
+        foreach (MyRadio_Season::getAllSeasonsInLatestTerm() as $season) {
+            foreach ($season->getAllTimeslots() as $timeslot) {
+                if ($timeslot->getStartTime() < time()
+                    && $timeslot->getStartTime() > time() - 604800
+                ) {
+                    foreach ($timeslot->getSigninInfo() as $info) {
+                        if (isset($info["location"]) && $info["location"] != $no_track) {
+                            if (isset($info["user"])) {
+                                $eduroam = $info["user"]->getEduroam();
```
markspolakovs, line 48: People without eduroams exist. Idk if any will want to use the studios though, but is there a reason you used getEduroam instead of getEmail?

src/Classes/Daemons/MyRadio_TrackAndTraceDaemon.php
```diff
@@ -0,0 +1,97 @@
...
+        usort($data, function ($a, $b) {
+            if ($a["unix"] == $b["unix"]) {
+                return 0;
+            }
+            return ($a["unix"] < $b["unix"]) ? -1 : 1;
+        });
+
+        foreach ($data as $row) {
+            $table .= "<tr><td>" . $row["type"]
+                . "</td><td>" . $row["info"]
```
markspolakovs, line 83: If it's a guest, `$row["info"]` will be a PHP array, I haven't tried it but I suspect it won't format nicely.

src/Classes/ServiceAPI/MyRadio_Season.php
```diff
@@ -302,7 +302,19 @@ public static function create($params = [])
...
 
-        return self::getInstance($season_id);
+        $newSeason = self::getInstance($season_id);
+
+        // COVID Studio Usage - Summer 2021
+        if (isset($params['studio-request']) && $params['studio-request']) {
+            MyRadioEmail::sendEmailToList(
+                MyRadio_List::getByName("programming"),
+                "Studio Usage Request",
+                "Season " . $season_id . " of " . $newSeason->getMeta("title")
```
markspolakovs, line 312: This will be something like `Season 85939 of Some Show`, rather than `Season 3 of Some Show`. Is this intented?

markspolakovs (changes requested):
Quite a few of these are "I'm not sure if this is really what you meant" - if it is, happy to change this to approve.

src/Classes/ServiceAPI/MyRadio_Season.php
michael-grace: yes

src/Classes/Daemons/MyRadio_TrackAndTraceDaemon.php
michael-grace: the effort of dealing with that in a general case to the likelihood of needing to worry about it in an individual case isn't worth dealing with 😜

src/Classes/Daemons/MyRadio_TrackAndTraceDaemon.php
michael-grace: imma blame you if it don't work :)

src/Classes/Daemons/MyRadio_TrackAndTraceDaemon.php
markspolakovs: Just tried it, and it doesn't even work:

```
php > echo "<td>" . ["html" => "testytest"] . "</td>";
PHP Notice:  Array to string conversion in php shell code on line 1
<td>Array</td>
```

src/Classes/Daemons/MyRadio_TrackAndTraceDaemon.php
michael-grace: some people have personal emails set, but cause yusu/uni we need eduroams.

src/Classes/Daemons/MyRadio_TrackAndTraceDaemon.php
michael-grace: 🤔

src/Classes/Daemons/MyRadio_TrackAndTraceDaemon.php
```diff
@@ -0,0 +1,94 @@
...
+                                    "time" => CoreUtils::happyTime($info["time"]),
+                                    "unix" => $info["time"]
+                                ];
+                            } elseif ($info["guest_info"]) {
+                                $data[] = [
+                                    "type" => "Guest",
+                                    "info" => [
+                                        "display" => "html",
+                                        "html" => nl2br($info["guest_info"])
+                                    ],
```
markspolakovs, line 62:
```suggestion
                                    "info" => nl2br($info["guest_info"]),
```

src/Classes/Daemons/MyRadio_TrackAndTraceDaemon.php
michael-grace: lol

src/Classes/Daemons/MyRadio_TrackAndTraceDaemon.php
michael-grace: i mean, lol exceptions

markspolakovs (approved):

PR #1000, opened 2021-04-12
PR #1000

michael-grace:
Realistically, I did this just to be PR #1000. We don't actaully need to merge this until we're ready to 😛

### DIFF
--- a/src/Menus/menu.json
+++ b/src/Menus/menu.json
@@ -128,9 +128,6 @@
                             "title": "WebStudio",
                             "url": "module=MyRadio,action=webstudio",
                             "description": "Access WebStudio. Plan, create, practise and broadcast your shows!"
-                        },
-                        {
-                            "template": "MyRadio/showPlanner.twig"
                         }
                     ]
                 },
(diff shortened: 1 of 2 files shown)


mstratford (approved):
It's time.

src/Menus/menu.json
```diff
@@ -118,9 +118,6 @@
                             "title": "WebStudio",
                             "url": "module=MyRadio,action=webstudio",
                             "description": "Access WebStudio. Plan, create, practise and broadcast your shows!"
-                        },
-                        {
-                            "template": "MyRadio/showPlanner.twig"
```
markspolakovs, line 123: Should we delete this template, so we don’t have loose files lying around?

mstratford: There's a load of dead showplanner templates etc that could be chucked. Back from the days when it was attempted to control BAPS2

markspolakovs: Eh, I'll file a separate issue for that to not block this.

PR #1002, opened 2021-04-20
Add mapping of tracklist sources to sel sources

mstratford:
This is to allow support via API for BAPS3 to tracklist always, but only show in tracklists if the studio is on air.

Instead of using the server name mapping in BAPS2, this proposes to create 3 tracklist source chars (studio red, blue and OB), configured on each server.

Note, the table will need to be created (if I haven't done it already) in prod db, and all source mapping combos added (including webstudio to sel 5, etc).

Needs testing.

### DIFF
--- a/schema/base.sql
+++ b/schema/base.sql
@@ -2112,7 +2112,7 @@ CREATE TABLE selsources (
     selaction integer NOT NULL,
     sourceid character(1) NOT NULL
 );
-COMMENT ON TABLE selbaps IS 'Marries selector actions with tracklist sources';
+COMMENT ON TABLE selsources IS 'Marries selector actions with tracklist sources';
 CREATE TABLE source (
     sourceid character(1) NOT NULL,
     source text NOT NULL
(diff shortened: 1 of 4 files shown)


schema/base.sql
```diff
@@ -2108,6 +2108,11 @@ CREATE TABLE selbaps (
     bapsloc integer NOT NULL
 );
 COMMENT ON TABLE selbaps IS 'Marries selector actions with BAPS servers';
+CREATE TABLE selsources (
```
markspolakovs, line 2111: Any reason you're using base.sql rather than patches? Those are easier to apply on prod.

mstratford: oh, we're using them now? Ok :)

mstratford:
So it generally works.

schema/base.sql
mstratford: Done.

michael-grace (approved):

PR #1004, opened 2021-05-02
Remove duplicate credits from credit string.

mstratford:
Not yet tested.

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Creditable.php
+++ b/src/Classes/ServiceAPI/MyRadio_Creditable.php
@@ -108,16 +108,19 @@ trait MyRadio_Creditable
                 $credit_types_in_byline[] = $type["value"];
             }
         }
-        $str = '';
+        $names = [];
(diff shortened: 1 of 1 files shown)


MattWindsor91:
Looks like a universal improvement to me, but I don't know much PHP these days 😅

src/Classes/ServiceAPI/MyRadio_Creditable.php
```diff
@@ -108,16 +108,19 @@ public function getPresenterString()
...
             } else {
                 continue;
             }
         }
+        $names = array_unique($names);
+        $str = implode(", ",$names);
+        
 
-        return empty($str) ? '' : substr($str, 0, -2);
+        return empty($names) ? '' : $str;
```
MattWindsor91, line 123: Is there any reason why this can't be `empty($names) ? '' : implode(", ",$names)`?  Alternatively, what happens when `implode` is called on an empty array - if it's just `''`, could this ternary be eliminated?

PR #1007, opened 2021-05-07
Mark podcast as dirty when done converting

markspolakovs:
This _might_ fix the "need to yeet cache after uploading" issues. I'd quite like to make a nicer "pending state" implementation rather than just assuming `{file, submitted} == null => pending`, but that's outside the scope of this PR.


markspolakovs:
Nope, this won't work :(

PR #1011, opened 2021-05-09
Archiving Playlists

michael-grace:
cause deleting playlists is an absolute nightmare becuase of all the foreign keys and stuff, but music team still want to be able to yeetanise stuff


markspolakovs:
So this works (I assume)... but why did you add a separate page, instead of a parameter on the existing listPlaylists page?

Joseph-S12:

PR #1015, opened 2021-05-16
MyRadio_Podcast::getPodgetPodcastsAttachedToUser: fix API

markspolakovs:
The API thinks a MyRadio_User param is an int. It is not.

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Podcast.php
+++ b/src/Classes/ServiceAPI/MyRadio_Podcast.php
@@ -174,11 +174,11 @@ class MyRadio_Podcast extends MyRadio_Metadata_Common
     /**
      * Get all the Podcasts that the User is Owner of Creditor of.
      *
-     * @param MyRadio_User $user Default current user.
+     * @param int|MyRadio_User $user Default current user.
      *
      * @return MyRadio_Podcast[]
      */
-    public static function getPodcastsAttachedToUser(MyRadio_User $user = null)
+    public static function getPodcastsAttachedToUser($user = null)
     {
         return self::resultSetToObjArray(self::getPodcastIDsAttachedToUser($user));
     }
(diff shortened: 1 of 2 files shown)


markspolakovs:
Tested working on staging with the latest commit. To try it yourself, go to https://ury.org.uk/api-staging/v2/podcast/podcastsattachedtouser?api_key=YOUR_API_KEY&user=AN_USER_ID

markspolakovs:
Wait, hang on, it's misbehaving.

markspolakovs:
Right, that should work.

markspolakovs:
r?'ing @mstratford as this now touches some MyRadio internals and I'd appreciate an extra pair of eyes

src/Classes/ServiceAPI/MyRadio_Podcast.php
```diff
@@ -174,26 +174,32 @@ protected function __construct($podcast_id)
...
+     * @param int|string|MyRadio_User $user Default current user.
      *
      * @return int[]
      */
-    public static function getPodcastIDsAttachedToUser(MyRadio_User $user = null)
+    public static function getPodcastIDsAttachedToUser($user = null)
     {
         if ($user === null) {
             $user = MyRadio_User::getInstance();
+        } elseif (is_int($user)) {
```
mstratford, line 197: May just be GitHub mobile, but is the indentation correct here?

mstratford:
I've never been 100% sure how the mixins work, I'd appreciate finding out why the mixins come into these changes?

mstratford (commented):
Seems generally fine, but why mixins?

markspolakovs:
@mstratford Mixins are only ever used in `ServiceAPI::toDataSource()` to customise what fields you want (e.g. compare https://ury.org.uk/api/v2/user/currentuser and https://ury.org.uk/api/v2/user/currentuser?mixins=training,shows,personal_data,all_officerships).

They shouldn't be finding their way into any other methods, yet when I used this branch as of the commit-before-last with a mixin (e.g. https://ury.org.uk/api-staging/v2/podcast/podcastsattachedtouser?api_key=YOUR_API_KEY&user=AN_USER_ID&mixins=show) I got an error about the `$user` parameter - sticking in a var_dump revealed that it was passing in the value of the `mixins` field into the `$user` parameter. The simplest fix was to filter it out like we already do with `api_key`.

src/Classes/ServiceAPI/MyRadio_Podcast.php
markspolakovs: Looks fine to me

PR #1016, opened 2021-05-17
Support Links

michael-grace:


markspolakovs (approved):
Pulled into staging, and it does work (once I added in a GLOBAL_ACCESS permission) - I hate the fact that you used tabledata, but it is effective (though I'm not sure we want there to be a search bar on a page with six entries...)

michael-grace:
my plan was to put a no login auth on it, just cause we want the sharing gainz

markspolakovs:
@michael-grace tbh in that case I'd personally prefer if it were on 2016-site, because I don't think there's precedent for publicly-human-visible MyRadio links (and it's a bit ugly imo)

SplashyFish14:
I don't love this, but I suppose it's fine for now, would it be possible for it to be easily editable without doing a PR?

SplashyFish14:
Ok, after more thonking, thoughts on having the page on the wiki and just linking to it for now? saves you writing bad code :P

michael-grace:
...wonders why we didnt just think of a wiki page originally...


And its not **_bad_** code

markspolakovs:
@michael-grace @SplashyFish14 can this be closed?

PR #1022, opened 2021-08-21
Short URLs

markspolakovs:
Goes in tandem with https://github.com/UniversityRadioYork/2016-site/pull/297

### DIFF
--- a/schema/patches/10.sql
+++ b/schema/patches/10.sql
@@ -1,0 +1,17 @@
+BEGIN;
+
+CREATE TABLE public.short_urls (
+    short_url_id SERIAL PRIMARY KEY,
+    slug TEXT NOT NULL,
+    redirect_to TEXT NOT NULL
+);
+
+CREATE TABLE public.short_url_clicks (
+    click_id BIGSERIAL PRIMARY KEY,
+    short_url_id INTEGER REFERENCES public.short_urls (short_url_id) ON DELETE CASCADE,
+    click_time TIMESTAMPTZ NOT NULL,
+    user_agent TEXT DEFAULT NULL,
+    ip_address INET DEFAULT NULL
+);
+
+COMMIT;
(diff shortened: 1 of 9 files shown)


schema/patches/10.sql
```diff
@@ -0,0 +1,19 @@
...
+
+CREATE TABLE public.short_url_clicks (
+    click_id BIGSERIAL PRIMARY KEY,
+    short_url_id INTEGER REFERENCES public.short_urls (short_url_id) ON DELETE CASCADE,
+    click_time TIMESTAMPTZ NOT NULL,
+    user_agent TEXT DEFAULT NULL,
+    ip_address INET DEFAULT NULL
+);
+
+INSERT INTO public.l_action (descr, phpconstant) VALUES ('Edit Short URLs', 'AUTH_EDITSHORTURLS');
```
michael-grace, line 17: could this be added in `schema/whichever` file so its with all the others, and all of those get added when someone makes a local myradio. and we can just add this one ourselves to prod

src/Classes/ServiceAPI/MyRadio_ShortURL.php
```diff
@@ -0,0 +1,196 @@
...
+            [
+                'title' => 'Edit Short URL'
+            ]
+        ))->addField(
+            new MyRadioFormField(
+                'slug',
+                MyRadioFormField::TYPE_TEXT,
+                [
+                    'label' => 'Slug',
+                    'explanation' => "The bit that goes after https://$domain. Don't include $domain or a slash."
```
michael-grace, line 143: does this/could this fill from the config `short_url` or whatever the fields called?

markspolakovs: I don't think we have one of those - might you be confusing it with `short_name`?

schema/patches/10.sql
markspolakovs: Oh yeah that exists, will change

src/Classes/ServiceAPI/MyRadio_ShortURL.php
michael-grace: uh.. we've got `base_url` but thats got the /myradio bit on.

michael-grace (approved):

PR #1027, opened 2021-09-19
change demoed (studio proficient or ws trained) to studio trained

michael-grace:
its safe to say, if your're studio trained you're studio proficient


Joseph-S12:

PR #1029, opened 2021-10-03
BREAKING CHANGE: createOrActivate return null when user exists

markspolakovs:
Every year we have the problem of people getting confused when they sign up and don't get an email, because they already had an account.

This commit changes MyRadio_User::createOrActivateUser() to return null when the user already existed before it was called.

This is a BREAKING CHANGE because, although all MyRadio callsites have been updated, various places use its API and would need updating too (off the top of my head, /getinvolved and the alumni signup form).

This also technically introduces an account enumeration vulnerability, but personally I think the UX benefit is worth the marginal added risk.


mstratford (approved):
LGTM

PR #1031, opened 2021-10-06
Add regex support for LDAP to block fake logins.

mstratford:
Looking for someone without chance of any of the other authenticators to login to test it.

### DIFF
--- a/src/Classes/Config.php
+++ b/src/Classes/Config.php
@@ -617,7 +617,7 @@ final class Config
     public static $auth_ldap_root = 'ou=people,ou=csrv,ou=nos,dc=york,dc=ac,dc=uk';
     public static $auth_db_user = '';
     public static $auth_db_pass = '';
-    public static $auth_ldap_regex = ''; // eg '/[a-z]{2,}[0-9]+(@york.ac.uk)?/'
+    public static $auth_ldap_regex = '/[a-z]{2,}[0-9]+(@york\.ac\.uk|)$/';
 
     /**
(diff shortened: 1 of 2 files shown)


src/Classes/MyRadio/MyRadioLDAPAuthenticator.php
```diff
@@ -37,6 +37,11 @@ public function __destruct()
      */
     public function validateCredentials($user, $password)
     {
+        # Check that it looks like a legit user first.
+        if (!empty(Config::$auth_ldap_regex) && !preg_match(Config::$auth_ldap_regex, $user)) {
+            sleep(0.01); # Emulate the speed of LDAP to help with timing attacks.
```
LordAro, line 42: Not sure this really helps with timing attacks. At least needs to be an random amount of sleeping, but I'd say the whole thing is necessary anyway

mstratford: Cheers :)

src/Classes/Config.php
```diff
@@ -617,6 +617,7 @@ final class Config
     public static $auth_ldap_root = 'ou=people,ou=csrv,ou=nos,dc=york,dc=ac,dc=uk';
     public static $auth_db_user = '';
     public static $auth_db_pass = '';
+    public static $auth_ldap_regex = ''; // eg '/[a-z]{2,}[0-9]+(@york.ac.uk)?/'
```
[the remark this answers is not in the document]
mstratford: Fair enough :)

PR #1032, opened 2021-10-06
UI fixes and niceties.

mstratford:
Just a growing collection of stuffs that would be nicer if...

- Say welcome instead of "Menu".
- Reset the selected timeslot when impersonating if the user doesn't have permission to access it
- Explain what an audio log is
- Remove non-manual uploads


markspolakovs:
@mstratford should this still be open?

mstratford:
Yeah, still look like valid changes to me :)

src/Templates/NIPSWeb/manage_library.twig
```diff
@@ -1,30 +0,0 @@
-<!DOCTYPE html>
```
markspolakovs, line 1: Should the associated controller be deleted as well?

markspolakovs (approved):

src/Templates/NIPSWeb/manage_library.twig
mstratford: I forget, if it's not the same as the manual upload training one, then yes.

mstratford: Doesn't look like it's actually in the code anymore, just the manual upload one is in there.

PR #1038, opened 2021-11-01
Set a variable attendee count for demos

cs-499243:
Adds the ability to set an amount of attendees for a training event, making the "max of 2" more variable. Useful for cases when either 1 person is wanted or an arbitrarily large amount of people are wanted. 

Also there are slight changes like making a function return early when it (not doing anything else) throws an error and testing if an edited event has a valid date. 

Warning: I have not tested this so I don't know if it will work.

### DIFF
--- a/schema/patches/6.sql
+++ b/schema/patches/6.sql
@@ -10,6 +10,7 @@ create table schedule.demo
 			references l_presenterstatus,
 	demo_time timestamp with time zone not null,
 	demo_link text,
+	num_people int,
 	memberid int
 		constraint demo_member_memberid_fk
 			references member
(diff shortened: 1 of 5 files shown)


michael-grace:
Looks ok, but I haven't tested it. You're right about removing the else's after throwing the exception, but returning isn't necessary as the exception is thrown

cs-499243:
That's fair - just playing it safe in that case. I'll try and move this along once we can run myradio locally again

markspolakovs:
Closing in favour of https://github.com/UniversityRadioYork/MyRadio/pull/1050.

PR #1039, opened 2021-11-13
Fix local install process

### LINKED ISSUE #1037: postgres-12 migration required for base.sql
Due to [changes in postgres 12](https://www.postgresql.org/docs/12/release-12.html#id-1.11.6.13.4) when you try to import the file [`base.sql`](../blob/master/schema/base.sql) you get `ERROR: tables declared WITH OIDS are not supported`. Because of this, [the setup](../blob/master/src/Controllers/Setup/dbschema.php#L44) just ignores the invalid code and moves past, meaning the database isn't built at all. Also not sure why the try-catch isn't doing much of anything around that.

`base.sql` has 27 instances of [`default_with_oids`](https://postgresqlco.nf/doc/en/param/default_with_oids/) in it so what can we do? We can't just not go to postgres-12 cause then the whole thing fails to run. Or at least I think that's the problem.

cs-499243:
- Do basic version updates for postgres and php
- Remove oids from base.sql - Closes #1037 
- Create basic bash script to wipe database cleanly (so you can edit the setup scripts without rerunning `vagrant up` each time)
- Set MYRADIO_CURRENT_SCHEMA_VERSION to the latest patch version (which I think we were meant to have been doing, since that's how it sees whether to install patches)
- Create a loop in dbschema because it has an issue where it applies base.sql but then, on the next query, doesn't seem to be able to find its changes in the database. This makes it repeat maybe 3 times but limits the amount a user has to do to get it working (which is a fair compromise)
- Other minor edits to dbschema to avoid weird unexpected cases

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,8 @@ CONNECT:
 DATABASE:
  - On the intro screen press "Click here to continue"
  - Enter the database details (see Database Credentials) and press Next
- - Press "run task"
+ - Press "run task", wait a few seconds and then press "run task" again.
+ - [This delay and repetition is a workaround for a slight bug in how we build the database]
  
 USER:
  - [Here you can make config changes but the defaults are autofilled]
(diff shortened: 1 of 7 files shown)


mstratford (commented):
Looking good, will try to test tomorrow :)

PR #1044, opened 2021-12-04
Add basic setup and db documentation

cs-499243:
- Clean up README by moving all the setup processes to one file in /docs
- Add explanations in the sql patches as to what the things do. Most of them are fairly simple but a summary lets you briefly scan through them and see what they do
- Add an explanation of how the database setup works, going through the different php pages and their assumptions.
- Added reference to where MYRADIO_CURRENT_SCHEMA_VERSION is defined


schema/patches/3.sql
```diff
@@ -42,7 +47,9 @@ AS $$
                       (effective_to >= NOW() OR effective_to IS NULL)
         );
 
-        -- Right, we've got what we need
+        -- Right, we've got what we need so dump output into analytics table
+	--  Preserve reference data (like session_id, current datetime) for comparison
+	--  TODO Explain why num_shows is rounded in that way
```
markspolakovs, line 52: Anonymity

cs-499243: Does nothing else in the analytics show which member it's referencing? Also what is this actually used for?

markspolakovs: The idea of the design is that an individual analytics record wouldn’t identify any one member (you can guess based on officerships, but for people without officerships all you get is the number of shows to the nearest 5). I put it in when I was planning myradio2 (lol) to see which pages what “kinds of people” used more. I never removed it - though it might be disabled on prod now.

cs-499243: Right that explains why it's not related to the stats pages then. I've added a brief bit to take note of that, even if it's not strictly being used. Is the rest of the PR alright?

docs/install.md
```diff
@@ -0,0 +1,138 @@
...
+USER:
+ - [Here you can make config changes but the defaults are autofilled]
+ - Press "complete starting set", scroll and press "save and continue"
+ - Input any first and last name, an email (NOT an @york.ac.uk email) and a password
+ - [If you enter an @york.ac.uk email you will not be able to login at all]
+ - Login using the email and password you just enterted
+
+## Default Credentials
+Database: (when building the database, these credentials are needed)
+ - Hostname: localhost
```
markspolakovs, line 87:
```suggestion
 - Hostname: `postgres` if running in Docker, `localhost` otherwise
```

schema/patches/2.sql
```diff
@@ -1,9 +1,10 @@
 BEGIN;
 
+-- Give the option for podcasts to be explicit (true or false)
```
markspolakovs, line 3:
```suggestion
-- Give the option for podcasts to be explicit (true or false).
-- This is used when generating RSS feeds for iTunes.
```

schema/patches/3.sql
```diff
@@ -25,14 +29,16 @@ AS $$
         officerships INTEGER[];
         num_shows INTEGER;
     BEGIN
-        -- Find officerships
+        -- Find the officerships this member currently has
+	--  Checks both analytics history and officers with this memberid
```
markspolakovs, line 33: Not quite - `create_analytics_record` is simply the name of the function, so all `create_analytics_record.memberid` does is reference the `memberid` parameter (prefixed with the function name otherwise `memberid` is ambiguous).

schema/patches/9.sql
```diff
@@ -1,3 +1,6 @@
+-- Link show tracklists to selector sources
+--  Moved from base.sql into a separate patch
+--  TODO Explain this better
```
markspolakovs, line 3: Used by BAPS and WebStudio for tracklisting (I think - @mstratford do you have the context for this?)

markspolakovs (approved):

schema/patches/9.sql
cs-499243: I imagine it's just so you can arbitrarily link tracklist sources to selector sources but, not knowing how selector works in that way, it's hard to briefly describe it more than the db comment.

PR #1050, opened 2022-01-27
Training cut-off + people limit (fixes #1049, fixes #1052)

### LINKED ISSUE #1049: Close training session joining some time before it starts
Exact time still needs bikeshedding, but the idea is to ensure that people know well in advance if someone will be there at their training session so they don't go to campus for no reason (or conversely stay home if they think nobody's coming).

### LINKED ISSUE #1052: Allow changing the number of people to train at a time

markspolakovs:
Make the cutoff configurable, so that people can set it accordingly - early morning sessions may want notice, while people who are going to be at the station anyway don'ŧ really mind.

Screenshots in action (on my local dev environment):
image
image
image

Also merged with @hicks927's work on setting an attendee limit, because it makes sense to do them at the same time:

image
image

### DIFF
--- a/schema/patches/12.sql
+++ b/schema/patches/12.sql
@@ -3,4 +3,7 @@ BEGIN;
 ALTER TABLE schedule.demo
     ADD COLUMN signup_cutoff_hours INT DEFAULT 0;
 
+ALTER TABLE schedule.demo
+    ADD COLUMN max_participants INT DEFAULT 2;
+
 COMMIT;
--- a/schema/patches/12.sql
+++ b/schema/patches/12.sql
@@ -1,0 +1,6 @@
+BEGIN;
+
+ALTER TABLE schedule.demo
+    ADD COLUMN signup_cutoff_hours INT DEFAULT 0;
+
+COMMIT;
--- a/src/Controllers/Training/leaveDemo.php
+++ b/src/Controllers/Training/leaveDemo.php
@@ -3,5 +3,5 @@
 use \MyRadio\MyRadio\URLUtils;
 use \MyRadio\ServiceAPI\MyRadio_Demo;
 
-MyRadio_Demo::getInstance($_REQUEST['demoid'])->leave();
-URLUtils::redirect($module, 'listDemos', ['msg' => 3]); // 3 means left... see listDemos.php
+$result = MyRadio_Demo::getInstance($_REQUEST['demoid'])->leave();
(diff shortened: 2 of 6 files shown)


src/Controllers/Training/leaveDemo.php
```diff
@@ -3,5 +3,5 @@
 use \MyRadio\MyRadio\URLUtils;
 use \MyRadio\ServiceAPI\MyRadio_Demo;
 
-MyRadio_Demo::getInstance($_REQUEST['demoid'])->leave();
-URLUtils::redirect($module, 'listDemos', ['msg' => 3]); // 3 means left... see listDemos.php
+$result = MyRadio_Demo::getInstance($_REQUEST['demoid'])->leave();
+URLUtils::redirect($module, 'listDemos', ['msg' => $result === 0 ? -1 : $result * -1]); // -1 means left... see listDemos.php
```
cs-499243, line 7: The -n messages are somewhat confusing. I see why you're doing it (to make leaving distinguished from joining) but if you do that maybe http status code style grouping is better (101, 201, 301 etc.)? Even then is it necessary? Especially when your only cases are -1 and -3.

schema/patches/12.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+
+ALTER TABLE schedule.demo
+    ADD COLUMN signup_cutoff_hours INT DEFAULT 0;
```
cs-499243, line 4: It would be nice if we could group sql patches together by topic. Could we remove 6.sql and merge it into 12.sql?

cs-499243 (approved):
Code looks fine - only comments I have are about choices of arranging things.

schema/patches/12.sql
markspolakovs: The idea behind patches was that we never go back on them - each patch is an "upgrade" of the previous version, otherwise we could have fun stuff happen in production. In theory we could remove older patches when they're superseded, however that makes my head hurt more than just keeping them around forever.

src/Controllers/Training/leaveDemo.php
markspolakovs: Yeah I wasn't exactly happy about this when I made it - my reasoning is that both 3 codes are for the time limit (+3 for joining and -3 for leaving). I think I'll stick to it like this, because so far there's not enough to justify finer grouping in my head.

michael-grace (approved):

michael-grace:
tho maybe put cutoff hours above optional zoom link

PR #1061, opened 2022-06-17
AutoViz control

markspolakovs:


michael-grace:
i mean, i'll assume this works

Have you considered a MyRadio_Daemon to make clips expire and delete them after some threshold (i.e. 10 weeks - can be configurable), so we're not collecting video files? Also, (though this isn't really part of the PR, just a comment), might be an idea to make the cllips directory a symlink to somewhere on the filestore, just cause that's where there's likely to be more space. (you can really tell storage management has been a fair chunk of my placement 😂)

markspolakovs:
@michael-grace 

> Have you considered a MyRadio_Daemon to make clips expire and delete them after some threshold (i.e. 10 weeks - can be configurable), so we're not collecting video files

Yes - not right now, but definitely before we launch it "for good" (i.e. next year).

> Also, (though this isn't really part of the PR, just a comment), might be an idea to make the cllips directory a symlink to somewhere on the filestore, just cause that's where there's likely to be more space. (you can really tell storage management has been a fair chunk of my placement 😂)

Already done - clips are stored in `/filestore/Shows/AutoViz` on Steve, NFS-mounted on thunderhorn and SMB-mounted on dicker.

src/Classes/ServiceAPI/MyRadio_AutoVizConfiguration.php
```diff
@@ -0,0 +1,161 @@
...
+    private int $show_season_timeslot_id;
+    private bool $record;
+    /** @type string|null */
+    private $stream_url;
+    /** @type string|null */
+    private $stream_key;
+
+    protected function __construct(array $data)
+    {
+                $this->autoviz_config_id = (int) $data['autoviz_config_id'];
```
markspolakovs, line 20:
```suggestion
        $this->autoviz_config_id = (int) $data['autoviz_config_id'];
```

PR #1063, opened 2022-08-14
Check selector to display currentAndNext off-air vs Jukebox

markspolakovs:


markspolakovs:
Verified on Staging:
image

michael-grace:
do we want to do it this way, or change the `isTerm` function this (and potentially other stuff) references??

Joseph-S12:
Given most of the summer is (this and potentially future years) a 'term' it is probably better to use this over isTerm.

markspolakovs:
> do we want to do it this way, or change the `isTerm` function this (and potentially other stuff) references??

actually nothing else uses `isTerm()`, lol (in myradio itself at least, not sure about the API).

Joseph-S12 (approved):
I can't really test it, but as HoC, I am happy for it to 'go live' if you have tested it.

PR #1064, opened 2022-08-15
Update enable disable message on autoViz table

ColinRoitt:
Make the enable disable feature more clear in its feature and function


mstratford (approved):
Seems legit

PR #1065, opened 2022-09-18
Event duplication

markspolakovs:
Replaces the recurring events feature (which was never fully implemented) with a button to duplicate a given event.


michael-grace:
does it work?

markspolakovs:
Probably, let me try after the rebase

markspolakovs:
It works.

michael-grace (approved):

PR #1081, opened 2022-10-04
Clean up demo_attendee records for missed demos

### LINKED ISSUE #1080: if your session is in the past, and you aren't marked as trained, it suggests you missed the session, so you should be able to sign up for another instead of being pranked by myradio

markspolakovs:
Fixes #1080.


michael-grace (commented):
This could remove the use of the 'mark people attending this demo as trained in this type' button (not that I think anyone uses it anyway instead just going onto peoples profiles), as it could remove them from the demo during it, so they aren't on it at the end for that button to be any use. Maybe add an hour onto the start of demo time before they can be removed from it by this.

michael-grace (approved):

PR #1087, opened 2022-10-10
Cancel demos

### LINKED ISSUE #766: Add ability to delete a training slot.

markspolakovs:
Fixes #766.


michael-grace (approved):

michael-grace:
if it doesnt work and/or breaks myradio, it's not my fault

PR #1097, opened 2022-10-22
Don't send demo change emails unless anything meaningful changes

### LINKED ISSUE #1079: Only email training attendees when the time changes
Currently we email all training attendees whenever there is a change to the session, even if the time doesn't change, which can result in strange emails like

```
Hi Marks

There's been a change to your training session on 2022-10-04 15:15:00+01.

It is now at 04/10/2022 15:15.
```

We should check if the time has changed and only send emails then.

markspolakovs:
Fixes #1079.


michael-grace:
Does it work?

Dunno.

michael-grace:
It broke.

michael-grace:
It didnt work

PR #976, opened 2021-01-30
editing seasons

michael-grace:
oh...and a small issue in the setup. but mainly metadata fun

PR #1104, opened 2023-01-04
Hide "Edit Season" link

markspolakovs:
It hasn't worked in years, so let's avoid people clicking it and wondering why things broke.


michael-grace:
istr sorting this page like a year ago, tho it could have been a different page

michael-grace:
#976 unless there's other issues

markspolakovs:
appears to still be something d0dgy today - error report from earlier:
image

PR #1106, opened 2023-01-23
Fix double space in "Attending Training" email

markspolakovs:


michael-grace (dismissed):
So, you've just decided we are against double spaces... It should be a triple space

PR #1112, opened 2023-04-10
Wrap setShowPhoto in a transaction

### LINKED ISSUE #1111: MyRadio_Show::setShowPhoto should do its DB operations in a transaction
Today it first runs the `INSERT INTO show_image_metadata`, then runs `getimagesize` on the file:

https://github.com/UniversityRadioYork/MyRadio/blob/733458714abf756d4e5d35553da37fe20c9839b2/src/Classes/ServiceAPI/MyRadio_Show.php#L741-L772

If the image is invalid, it'll still have created the show_image_metadata record with garbage contents.

It should probably be in a transaction so it'd get rolled back if the image is invalid.

markspolakovs:
Fixes #1111.

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Show.php
+++ b/src/Classes/ServiceAPI/MyRadio_Show.php
@@ -738,6 +738,8 @@ class MyRadio_Show extends MyRadio_Metadata_Common
     */
     public function setShowPhoto($tmp_path)
     {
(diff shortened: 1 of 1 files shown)


JamboPE (approved):
Makes sense in theory, no idea if it'll work as I don't understand php yet but yolo

PR #1117, opened 2023-05-10
this years members on api

michael-grace:


markspolakovs (approved):
Seems fine, though not sure what the use case is :)

michael-grace:
turns out it's actually just already got merged in

PR #1125, opened 2023-08-06
fix userSelectedTimeslot

michael-grace:
because of `setUserSelectedTimeslot` being static (because it sets it in the session), it considers it the GET method.

By pure chance in the last 2.5 years, when going through the methods, it looked at setUST first, said it was the get method, then saw getUST and overwrote that as the GET method, but now it's seeing them in the other order, so overwrites getUST with setUST as the GET method. Because there's no ordering to the methods [I assume], just the order they happen to be returned in, and we've been relying on getUST coming second.

I put it in as an edge case cause it kinda is - static method setting a thing essentially 'globally'


markspolakovs (approved):
Good catch! Pulled into myradio-staging and it works.

I'd perhaps consider leaving a comment explaining the edge case so future readers don't have to find the commit message, but I'd be happy for this to be merged as is.

PR #1126, opened 2023-08-09
week names

michael-grace:
is there a better way? yes.
could i be bothered? no.
does it work? yes.


schema/patches/17.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE IF EXISTS public.terms
+    ADD COLUMN week_names text;
```
markspolakovs, line 2: May be better to store it as `JSONB` (or even `text[]`) to make the nature clearer? idk

src/Classes/ServiceAPI/MyRadio_Scheduler.php
```diff
@@ -194,6 +194,21 @@ public static function getTermWeeks($termid)
...
+
+	i.e. ['Week 0 Sem 1', 'Week 1 Sem 1'] etc
+
+	@param int $termid id of the term
+	@return array of strings, the week names
+    }
+    */
+    public static function getTermWeekNames($termid)
+    {
+	return json_decode(self::$db->fetchOne("SELECT week_names FROM terms WHERE termid = $1", [$termid])["week_names"]);
```
markspolakovs, line 209: Worth caching this? How often would `getTermWeekNames` be called?

JP5457:
if the term names need to be manually entered into the DB it could be nice to have some sort of message on the make new term page so that whoever is making a term knows to do that and how to format it. (Could also be good to mention the term start time needing to be set manually)

PR #1128, opened 2023-09-26
Jp jukebox activesong

JP5457:
Current jukebox playlist display on timelord

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Timeslot.php
+++ b/src/Classes/ServiceAPI/MyRadio_Timeslot.php
@@ -688,7 +688,7 @@ class MyRadio_Timeslot extends MyRadio_Metadata_Common
         $timeslot = self::getCurrentTimeslot($time, $filter);
         $next = self::getNextTimeslot($time, $filter);
 
-        $currentJukebox = iTones_playlist::getCurrentPlaylist();
+        $currentJukebox = iTones_playlist::getPlaylistAtTime(date('H:i:s'));
 
         //Still display a show if there's one scheduled for whatever reason.
         if (empty($timeslot)) {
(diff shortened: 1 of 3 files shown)


src/Classes/ServiceAPI/MyRadio_Timeslot.php
```diff
@@ -687,6 +688,8 @@ public static function getCurrentAndNext($time = null, $n = 1, $filter = [1])
         $timeslot = self::getCurrentTimeslot($time, $filter);
         $next = self::getNextTimeslot($time, $filter);
 
+        $currentJukebox = iTones_playlist::getPlaylistAtTime(date('H:i:s'));
```
markspolakovs, line 691:
```suggestion
        $currentJukebox = iTones_Playlist::getPlaylistAtTime(date('H:i:s'));
```

src/Classes/iTones/iTones_Playlist.php
```diff
@@ -235,6 +235,27 @@ public function getTitle()
         return $this->title;
     }
 
+    public function getPlaylistAtTime($time)
+    {
+        $getplaylist = self::$db->fetchColumn(
+        'SELECT playlist.playlistid
+        FROM jukebox.playlists playlist
+        INNER JOIN jukebox.playlist_availability avail ON playlist.playlistid=avail.playlistid
+        WHERE playlist.category=2
```
markspolakovs, line 244: Do we have a `Config` variable for the category ID? We'd rather not hard-code it if possible.

src/Classes/iTones/iTones_Playlist.php
```diff
@@ -235,6 +235,27 @@ public function getTitle()
         return $this->title;
     }
 
+    public function getPlaylistAtTime($time)
```
markspolakovs, line 238: What will this do if there are multiple active playlists at a time?

JP5457: just return one of them. This shouldn't matter as we are planning to move to only having a single active playlist at any point.

PR #1132, opened 2023-10-06
show art button

JP5457:
Adds a button that allows users to upload show art if they have permissions to edit that show


michael-grace (approved):
James says it works

PR #1134, opened 2023-10-22
GDPR compliance

JP5457:
three main features:
-A mandatory privacy policy that users must accept on their first attempt to login
-the ability to opt into and out of account deletion
-php scripts for emailing users eligible for deletion and for deleting users

### DIFF
--- a/scripts/gdprdeleteall.php
+++ b/scripts/gdprdeleteall.php
@@ -1,0 +1,49 @@
+#!/usr/local/bin/php -q
+<?php
+/**
+ * DO NOT RUN THIS SCRIPT WILLY NILLY
+ * this script will delete the personal data of any members that have been emailed regarding deletion and have not acted
+ * you should probably run the gdpremail script first then wait a week or two
+ * 
+ * run with "php gdprdeleteall.php"
+ */
+
+use \MyRadio\Database;
+use \MyRadio\MyRadioException;
+use \MyRadio\MyRadioError;
+use \MyRadio\MyRadioEmail;
+use \MyRadio\MyRadio\CoreUtils;
+use \MyRadio\ServiceAPI\MyRadio_User;
+
+ini_set('log_errors', 1);
+ini_set('error_log', '/tmp/php-gdpr-deleteall.log');
+ini_set('display_errors', 'On');
+
+require_once '../src/Controllers/root_cli.php';
+
+$db = Database::getInstance();
+
+$time = strtotime("-1 year", time());
+$date = date("Y-m-d", $time);
+
+echo "This script will delete details of all users that have not logged in for over a year\n Are you sure you want to continue? (y/n)";
+$cmdinput = trim(fgets(STDIN));
+if($cmdinput != 'Y'){
+    return;
+}
+echo "deleting user data\n";
+
+$db->query(
+    'UPDATE public.member
+	SET data_removal=\'default\'
+	WHERE data_removal=\'informed\' and last_login >= $1 ',
+    [$date]
+);
+
+$db->query(
+    'UPDATE public.member
+	SET college=10, phone=DEFAULT, receive_email=false, endofcourse=DEFAULT,  wheelchair=DEFAULT, data_removal=\'deleted\'
+	WHERE data_removal=\'informed\'',
+    []
+);
+?>
--- a/scripts/gdprdeleteuser.php
+++ b/scripts/gdprdeleteuser.php
@@ -1,0 +1,122 @@
+#!/usr/local/bin/php -q
+<?php
+/**
+ * DO NOT RUN THIS SCRIPT WILLY NILLY
+ * This script will completely yoink a users personal details, show credits, podcast credits and a bunch of other stuff.
+ * Only run this script if a user has explicity declared they wish to be forgotten and understand the consequences.
+ * 
+ * run with "php gdprdeleteuser.php [userid]"
+ */
+
+use \MyRadio\Database;
+use \MyRadio\MyRadioException;
+use \MyRadio\MyRadioError;
+use \MyRadio\MyRadioEmail;
+use \MyRadio\MyRadio\CoreUtils;
+use \MyRadio\ServiceAPI\MyRadio_User;
+
+ini_set('log_errors', 1);
+ini_set('error_log', '/tmp/php-gdpr-deleteuser.log');
+ini_set('display_errors', 'On');
+
+require_once '../src/Controllers/root_cli.php';
+
+$db = Database::getInstance();
+
+$time = strtotime("-1 year", time());
+$date = date("Y-m-d", $time);
+
+$deletedUserId = 1350883;
+
+$userid = $argv[1];
+
+echo "This script will delete all the users personal data and some database links to the user\n Are you sure you want to continue? (y/n)";
+$cmdinput = trim(fgets(STDIN));
+if($cmdinput != 'Y'){
+    return;
+}
+echo "User selected for deletion: " . $userid . "\n";
+
+try{
+    $db->query(
+        'INSERT INTO public.member(
+            memberid, fname, sname, college, receive_email, data_removal)
+            VALUES ($1, \'deleted\', \'user\', 10, false, \'deleted\')',
+        [$deletedUserId]
+    );
+}  catch (exception $e) {
+    echo 'deleting user\n';
+}
+
+$db->query(
+    'UPDATE schedule.show_credit SET memberid=$1 WHERE memberid=$2',
+    [$deletedUserId,$userid]
+);
+
+$db->query(
+    'UPDATE uryplayer.podcast_credit SET creditid=$1 WHERE creditid=$2',
+    [$deletedUserId,$userid]
+);
+
+$db->query(
+    'UPDATE bapsplanner.managed_items SET memberid=$1 WHERE memberid=$2',
+    [$deletedUserId,$userid]
+);
+
+$db->query(
+    'UPDATE schedule.timeslot_metadata SET memberid=$1 WHERE memberid=$2',
+    [$deletedUserId,$userid]
+);
+
+$db->query(
+    'UPDATE public.mail_alias_member SET memberid=$1 WHERE memberid=$2',
+    [$deletedUserId,$userid]
+);
+
+$db->query(
+    'UPDATE public.member_year SET memberid=$1 WHERE memberid=$2',
+    [$deletedUserId,$userid]
+);
+
+$db->query(
+    'UPDATE public.member_presenterstatus SET memberid=$1 WHERE memberid=$2',
+    [$deletedUserId,$userid]
+);
+
+$db->query(
+    'UPDATE public.member_pass SET memberid=$1 WHERE memberid=$2',
+    [$deletedUserId,$userid]
+);
+
+$db->query(
+    'UPDATE uryplayer.podcast_metadata SET memberid=$1 WHERE memberid=$2',
+    [$deletedUserId,$userid]
+);
+
+$db->query(
+    'UPDATE mail.email_recipient_member SET memberid=$1 WHERE memberid=$2',
+    [$deletedUserId,$userid]
+);
+
+$db->query(
+    'UPDATE uryplayer.podcast SET memberid=$1 WHERE memberid=$2',
+    [$deletedUserId,$userid]
+);
+
+$db->query(
+    'UPDATE public.mail_subscription SET memberid=$1 WHERE memberid=$2',
+    [$deletedUserId,$userid]
+);
+
+$db->query(
+    'UPDATE mail.alias_member SET destination=$1 WHERE destination=$2',
+    [$deletedUserId,$userid]
+);
+
+$db->query(
+    'UPDATE public.member
+	SET college=10, phone=DEFAULT, email=DEFAULT, receive_email=false, local_name=DEFAULT, local_alias=DEFAULT, account_locked=true, last_login=DEFAULT, endofcourse=DEFAULT, eduroam=DEFAULT, usesmtppassword=DEFAULT, joined=DEFAULT, require_password_change=DEFAUlT, profile_photo=DEFAULT, bio=DEFAULT, auth_provider=DEFAULT, contract_signed=DEFAULT, gdpr_accepted=DEFAULT, wheelchair=DEFAULT, data_removal=\'deleted\'
+	WHERE memberid=$1',
+    [$deletedUserId,$userid]
+);
+?>
--- a/scripts/gdpremail.php
+++ b/scripts/gdpremail.php
@@ -1,0 +1,81 @@
+#!/usr/local/bin/php -q
+<?php
+/**
+ * DO NOT RUN THIS SCRIPT WILLY NILLY
+ * this script will email literally everyone that has not logged into myradio in over a year. That is a lot of people.
+ * after running this and waiting a couple weeks you should probably run the gdprdelete script
+ * 
+ * run with "php gdpremail.php"
+ * 
+ * Not actually tested for obvious reasons but probably works
+ */
+
+use \MyRadio\Database;
+use \MyRadio\MyRadioException;
+use \MyRadio\MyRadioError;
+use \MyRadio\MyRadioEmail;
+use \MyRadio\MyRadio\CoreUtils;
+use \MyRadio\ServiceAPI\MyRadio_User;
+
+ini_set('log_errors', 1);
+ini_set('error_log', '/tmp/php-gdpr-email.log');
+ini_set('display_errors', 'On');
+
+require_once '../src/Controllers/root_cli.php';
+
+$db = Database::getInstance();
+
+$warning_email = <<<EOT
+<p>You are getting this email because you have not logged into MyRadio in over a year</p>
+<p>In one week all of your personally identifiable data that is not required for webstudio or our public facing websites to function as advertised will be deleted.</p>
+<p>If you wish to avoid this you can opt out of deletion by logging into your <a href="https://ury.org.uk/myradio">myradio account or by contact the ury computing team.</p>
+<p>If you are happy for your personal data to be deleted feel free to ignore this eamil.</p>
+--<br/>
+The URY Computing team<br/>
+<br/>
+University Radio York 1350AM 88.3FM<br/>
+---------------------------------------------<br/>
+<a href="mailto:head.of.computing@ury.org.uk">head.of.computing@ury.org.uk</a><br/>
+---------------------------------------------<br/>
+On Air | Online | On Tap<br/>
+<a href="https://ury.org.uk">ury.org.uk</a>
+EOT;
+
+$time = strtotime("-1 year", time());
+$date = date("Y-m-d", $time);
+
+echo "This script will Email all users that have not logged in for over a year\n Are you sure you want to continue? (y/n)";
+$cmdinput = trim(fgets(STDIN));
+if($cmdinput != 'Y'){
+    return;
+}
+echo "Emailing users\n";
+
+$memebersToEmail = $db->fetchAll(
+    'SELECT memberid, last_login
+	FROM public.member WHERE last_login <= $1 and joined <= $1',
+    [$date]
+);
+
+$db->query(
+    'UPDATE public.member
+	SET data_removal=\'informed\'
+	WHERE data_removal=\'default\' and last_login <= $1 and joined <= $1',
+    [$date]
+);
+
+$db->query(
+    'UPDATE public.member
+	SET data_removal=\'informed\'
+	WHERE data_removal=\'default\' and last_login IS NULL',
+    []
+);
+
+foreach($memebersToEmail as $member){
+    MyRadioEmail::sendEmailToUser(
+        $member["memberid"],
+        'MyRadio account deletion',
+        $warning_email
+    );
+}
+?>
--- a/src/Classes/ServiceAPI/MyRadio_User.php
+++ b/src/Classes/ServiceAPI/MyRadio_User.php
@@ -204,6 +204,30 @@ class MyRadio_User extends ServiceAPI implements APICaller
     private $contract_signed;
 
     /**
+     * True if user has agreed to our privacy statement.
+     *
+     * @var bool
+     */
+    private $gdpr_accepted;
+
+    /**
+     * default if user can be deleted
+     * informed if user has been warned about deletion
+     * optout if user has opted out of deletion
+     * deleted if user has been deleted
+     *
+     * @var enum
+     */
+    private $data_removal;
+
+        /**
+     * If true, hides user data from API.
+     *
+     * @var bool
+     */
+    private $hide_profile;
+
+    /**
      * Initiates the User variables.
      *
      * @param int $memberid The ID of the member to initialise
@@ -216,7 +240,8 @@ class MyRadio_User extends ServiceAPI implements APICaller
             'SELECT fname, sname, college AS collegeid, l_college.descr AS college,
             phone, email, receive_email::boolean::text, local_name, local_alias, eduroam,
             account_locked::boolean::text, last_login, joined, profile_photo, bio,
-            auth_provider, require_password_change::boolean::text, contract_signed::boolean::text
+            auth_provider, require_password_change::boolean::text, contract_signed::boolean::text, gdpr_accepted::boolean::text, 
+            data_removal, hide_profile::boolean::text
             FROM member, l_college
             WHERE memberid=$1
             AND member.college = l_college.collegeid
@@ -332,6 +357,11 @@ class MyRadio_User extends ServiceAPI implements APICaller
      */
     public function getAllTraining($ignore_revoked = false)
     {
+
+        if($this->isProfileHidden()){
+            return [];
+        }
+
         if (!$this->training) {
             // Get Training info all into array
             $this->training = self::$db->fetchColumn(
@@ -677,6 +707,10 @@ class MyRadio_User extends ServiceAPI implements APICaller
      */
     public function getOfficerships($includeMemberships = false)
     {
+        if ($this->isProfileHidden()){
+            return [];
+        }
+
         if (!empty($this->officerships) && !$includeMemberships) {
             return $this->officerships;
         }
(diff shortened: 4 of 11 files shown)


src/Templates/MyRadio/privacy.twig
```diff
@@ -0,0 +1,76 @@
...
+<ul>
+<li> To comply with legal obligations and resolve disputes.</li>
+<li> To provide required information to ofcom, which may include committee history.</li>
+<li> To protect the security and integrity of our website and users. </li>
+</ul>
+<h3> 3. Sharing of Information </h3>
+<p> We may share your information under the following circumstances: </p>
+<h4> a. With Third Parties: </h4>
+<ul>
+<li> We may share your information with third-party service providers who assist us in providing and improving our services. These providers are contractually obligated to protect your data. </li>
```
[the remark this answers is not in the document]
JP5457: I think its just covering us in case we want to do this

src/Templates/MyRadio/privacy.twig
```diff
@@ -0,0 +1,76 @@
...
+<p> 
+Welcome to MyRadio, the University Radio York members website! At URY, we value your privacy and are committed to protecting your personal information. This Privacy Policy outlines how we collect, use, disclose, and protect your data when you use our website.
+Please take a moment to read this Privacy Policy carefully to understand our practices regarding your personal information. By using the MyRadio website, you consent to the practices described herein.
+</p>
+<h3> 1. Information We Collect </h3>
+<p> We collect the following types of information when you use our website: </p>
+<h4> a. Personal Information: </h4>
+<ul>
+<li>When you register for an account, we collect your name, email address, and password.</li>
+<li>If you choose to provide it, we may collect additional information such as your college, profile picture, and other personal details.</li>
```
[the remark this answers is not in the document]
JP5457: The field can't be null be there is an unknown option which is basically the same thing?

src/Controllers/gdpremail.php
```diff
@@ -0,0 +1,68 @@
...
+
+ini_set('log_errors', 1);
+ini_set('error_log', '/tmp/php-mailparser-error.log');
+ini_set('display_errors', 'On');
+
+require_once __DIR__.'/root_cli.php';
+
+$db = Database::getInstance();
+
+$warning_email = <<<EOT
```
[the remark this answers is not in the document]
JP5457: It could but with the context of it only being sent out by a manually called php script I didn't really see the point of making it easily configurable.

src/Controllers/gdprdelete.php
```diff
@@ -0,0 +1,38 @@
...
+$db->query(
+    'UPDATE public.member
+	SET data_removal=0
+	WHERE data_removal=1 and last_login >= $1 ',
+    [$date]
+);
+
+$db->query(
+    'UPDATE public.member
+	SET college=10, phone=DEFAULT, email=DEFAULT, receive_email=DEFAULT, local_name=DEFAULT, local_alias=DEFAULT, account_locked=true, last_login=DEFAULT, endofcourse=DEFAULT, eduroam=DEFAULT, usesmtppassword=DEFAULT, joined=DEFAULT, require_password_change=DEFAUlT, profile_photo=DEFAULT, bio=DEFAULT, auth_provider=DEFAULT, contract_signed=DEFAULT, gdpr_accepted=DEFAULT, wheelchair=DEFAULT, data_removal=3
```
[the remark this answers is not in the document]
JP5457: Kinda, mostly its NULL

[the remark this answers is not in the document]
JP5457: I did check the sql runs properly

src/Classes/ServiceAPI/MyRadio_User.php
```diff
@@ -509,7 +526,7 @@ public function getPermissionsInfo()
     /**
      * Returns the User's email address. If the email address is null, it is
      * assumed their eduroam address is the preferred contact method.
-     *
+     *account_locked
```
LordAro, line 529: ?

src/Classes/MyRadio/MyRadioForm.php
```diff
@@ -291,7 +291,7 @@ public function editMode($ident, $values, $action = null)
      * Sets the values of this form to the given data. Note that this does not set the other fields needed to make it
      * an "edit" form - if you want that, you likely want {@link editMode}.
      * @param $values array
-     * @return self
+     * @return
```
LordAro, line 294: ?

src/Controllers/gdprdelete.php
```diff
@@ -0,0 +1,38 @@
...
+use \MyRadio\MyRadio\MyRadioForm;
+use \MyRadio\MyRadio\MyRadioFormField;
+use \MyRadio\MyRadio\MyRadioDefaultAuthenticator;
+use \MyRadio\ServiceAPI\MyRadio_User;
+
+ini_set('log_errors', 1);
+ini_set('error_log', '/tmp/php-mailparser-error.log');
+ini_set('display_errors', 'On');
+
+require_once __DIR__.'/root_cli.php';
```
LordAro, line 18: no other controllers have these, why is it here?

LordAro (commented):
Since I was in the area.

My primary concerns are conceptual though - are you sure that this is actually required? What's prompted adding an automatic deletion to MyRadio? Deletion requests should of course be followed, but what's collected currently in my view falls under "acceptable use" (except certain bits of personal data)

The URY database has such a wealth of knowledge in it regarding everything people did in their time at URY, going back over 20 years now.

Once it's gone it's gone, and it'd be a shame to delete anything unnecessarily - I wouldn't be surprised if this would delete 90% of the "members".

src/Classes/ServiceAPI/MyRadio_User.php
JP5457: I don't know either. I removed it and added but git was having none of it

LordAro: Define "having none of it"

src/Controllers/gdprdelete.php
JP5457: Requests are put through something using the databse i don't know. its safe tho.

src/Classes/ServiceAPI/MyRadio_User.php
```diff
@@ -203,6 +203,23 @@ class MyRadio_User extends ServiceAPI implements APICaller
...
+     *
+     * @var bool
+     */
+    private $gdpr_accepted;
+
+    /**
+     * 0 if user can be deleted
+     * 1 if user has been warned about deletion
+     * 2 if user has opted out of deletion
+     * 3 if user has been deleted
```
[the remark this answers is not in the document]
markspolakovs: or at least named members (eg `MyRadio_User::GDPR_STATUS_OPTED_OUT`)

src/Classes/ServiceAPI/MyRadio_User.php
```diff
@@ -216,7 +233,7 @@ protected function __construct($memberid)
             'SELECT fname, sname, college AS collegeid, l_college.descr AS college,
             phone, email, receive_email::boolean::text, local_name, local_alias, eduroam,
             account_locked::boolean::text, last_login, joined, profile_photo, bio,
-            auth_provider, require_password_change::boolean::text, contract_signed::boolean::text
+            auth_provider, require_password_change::boolean::text, contract_signed::boolean::text, gdpr_accepted::boolean::text, data_removal
```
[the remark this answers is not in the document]
markspolakovs: nowadays we do this through adding a file to patches/ (but yes, we do need one of those)

src/Controllers/gdpremail.php
```diff
@@ -0,0 +1,68 @@
...
+<p>You are getting this email because you have not logged into MyRadio in over a year</p>
+<p>In one week all of your personally identifiable data other than your name will be deleted.</p>
```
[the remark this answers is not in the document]
JP5457: I'd be very happy to delete names

src/Controllers/gdprdelete.php
michael-grace: Yeah, there's a folder for scripts, or should these be daemons anyway??

JP5457: is this not the scripts folder? I did check the script can't be run through get request.

LordAro (changes requested):
I do not like forcing this, but I'm seriously concerned by the direction this is going in, and I'd like some answers

Why are you doing this? What issue is being solved by deleting all 20 years worth of user data about shows and everything they did at URY?

LordAro:
I reiterate - once you implement this, you *cannot* undo it. You should be very very sure the data that is being deleted is only what's desired. Have you checked backups (and restoring from those backups) recently?

LordAro:
Since I can feel the inevitable "because GDPR" response, I'm going to point you at https://ico.org.uk/for-organisations/uk-gdpr-guidance-and-resources/data-protection-principles/a-guide-to-the-data-protection-principles/the-principles/storage-limitation/

> You can keep personal data for longer if you are only keeping it for public interest archiving, scientific or historical research, or statistical purposes.

Delete the bits that are not needed if you feel you must (I'm trying not to think about why `is_wheelchair` stored at all), but my view is that the database is primarily covered by the above and should not be touched

I feel the need to say that I'm not opposed to deletion in general - I'm the one that deleted the 'gender' field from the database after the thorny question of how that should be represented came up. The answer was of course that we didn't need it and so shouldn't be keeping it. And that was before GDPR...

adfw:
To put my tuppenth' worth in, since the initial GDPR regulation came about when I was Head of Computing in 2016 (e.g. it got passed and appeared upon the horizon), think a discussion was briefly held... (it certainly later led to the "why do we record gender" discussion as @LordAro mentions above).

Just to stick my nose in here too, I think there are several 'classes' of members you might want to consider before doing this:

1. Members who signed up but never presented a show or did anything substantive (e.g. signed up at freshers fair, maybe got studio trained but progressed no further)
2. Members who never presented a show but perhaps were on a team, had a committee position (perhaps rare, but certainly historically there were for e.g. a few people in the technical teams who never had a show)
3. Most importantly, members who were studio trained and presented shows

For 1., I fully agree that under GDPR those members details are fully deprecated, no longer need to be held, and there is no good reason to hold such data - agreed that given a grace period, their details should be removed.
For 2., I think there is a reasonable argument that if a member held a position of note in the society (e.g. was on committee), that the society should keep a record of such.
For 3., If a member in the database appeared in a show (there may be debate as to how 'important' they are - e.g. presenter, engineer, guest?), and given that URY has always been a broadcast radio station, those members should be fully aware that their participation and being a presenter (or similar) on a licensed broadcaster in the UK, then there is a reasonable 'public interest' in keeping a record, at a minimum, of their appearance on the schedule.

That does not mean that a Subject Access Request, or a Deletion Request, can't occur and those details be made non-public, but my view would be that by presenting on a broadcast radio station, you lose a certain amount of right to privacy at a very high level - e.g. show credit. Mixcloud recordings, photos etc. should of course be deleted on request.

I would also note that any email is unlikely to reach many people who didn't update their contact email address after leaving UoY, and their york.ac.uk account will now go nowhere.

I would certainly see if any advice is available from Ofcom on this in terms of retaining data about presenters beyond the statutory 48-day logs retention period - I would suggest contacting them for advice.

Happy Compliancing,
Anthony
URY, 2011-2016

LordAro:
Also come on dude, commit messages are there for a reason

LordAro:
Sounds suspiciously like an actual staging environment :o (though I might suggest actual dev work should be done on your local computer first)

Back in my day we just edited it directly on the server. Often production itself. *waves walking stick*

LordAro:
Still, it's not the notifications I'm bothered about - future you will thank past you if you actually write out what you're doing. Helps reduce the "why on earth is that there" factor that inevitably happens months (or weeks or days or hours) later

michael-grace:
This is why we do rebasing when it's ready. Myradio is difficult to work on locally, so we have testing versions on server. **We do care** about what commit messages end up in history.

JP5457:
I've made some significant changed to how this works:
- users now have the option of hiding their details in MyRadio
- automatic deletion now only deletes a few unnecessary columns
- script for deleting almost all data from users that explicitly request deletion

In the meeting we decided that any data visible on ury.org.uk we could defend keeping because it is required for the website to function as advertised.

src/Classes/ServiceAPI/MyRadio_User.php
```diff
@@ -203,6 +203,30 @@ class MyRadio_User extends ServiceAPI implements APICaller
...
+     *
+     * @var bool
+     */
+    private $gdpr_accepted;
+
+    /**
+     * deleted if user can be deleted
+     * informed if user has been warned about deletion
+     * optout if user has opted out of deletion
+     * deleted if user has been deleted
```
michael-grace, line 217: `deleted` here twice?

src/Classes/ServiceAPI/MyRadio_User.php
```diff
@@ -677,6 +707,10 @@ public function getAccountLocked()
      */
     public function getOfficerships($includeMemberships = false)
     {
+        if ($this->getHideProfile()){
+            return [];
+        }
+
```
michael-grace, line 713: i think not having these on myradio, instead the public website won't display the pages of people who've got hide profile, cos then you couldn't see it in myradio, or is that what we're going for?

src/Classes/ServiceAPI/MyRadio_User.php
```diff
@@ -796,6 +846,11 @@ public function getRequirePasswordChange()
      */
     public function getShows($show_type_id = 1, $current_term_only = false)
     {
+
+        if ($this->getHideProfile()){
+            return [];
+        }
+
```
michael-grace, line 853: same comment as above

src/Classes/ServiceAPI/MyRadio_User.php
```diff
@@ -1594,6 +1693,28 @@ public function getEditForm()
...
+                ]
+            )
+        );
+        $form->addField(
+            new MyRadioFormField(
+                'hide',
+                MyRadioFormField::TYPE_CHECK ,
+                [
+                'required' => false,
+                'label' => 'Hide profile on public website (Strongly discouraged for current members)',
```
michael-grace, line 1713: is it even an option for current members? say, "hide public profile on website after x amount of time from last show?"

src/Classes/ServiceAPI/MyRadio_User.php
```diff
@@ -2216,6 +2365,10 @@ public static function getBulkAddForm()
      */
     public function toDataSource($mixins = [])
     {
+        if ($this->getHideProfile()){
+            return $this->getEmptyData();
+        }
+
```
michael-grace, line 2371: ok, yeah i guess this is a way of doing it. maybe we should discuss whether this way, or just a parameter saying 'profileHidden': bool, so toDataSource could still be used?

scripts/gdpremail.php
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * DO NOT RUN THIS SCRIPT WILLY NILLY
+ * this script will email literally everyone that has not logged into myradio in over a year. That is a lot of people.
+ * after running this and waiting a couple weeks you should probably run the gdprdelete script
+ * 
+ * run with "php gdpremail.php"
+ * 
+ * Not actually tested for obvious reasons but probably works
+ */
+
```
michael-grace, line 12: should these be a daemon, so they run automatically and in a controlled environment, say once a month rather than a script with a warning at the top of it

scripts/gdpremail.php
```diff
@@ -0,0 +1,74 @@
...
+	WHERE data_removal=\'default\' and last_login <= $1 ',
+    [$date]
+);
+
+$db->query(
+    'UPDATE public.member
+	SET data_removal=\'informed\'
+	WHERE data_removal=\'default\' and last_login IS NULL ',
+    []
+);
```
michael-grace, line 65: yeah, but if someone signed up just before this, and haven't logged in, this'll include them, maybe have last_login is null and signed_up <= a year ago

scripts/gdprdeleteall.php
```diff
@@ -0,0 +1,42 @@
...
+require_once '../src/Controllers/root_cli.php';
+
+$db = Database::getInstance();
+
+$time = strtotime("-1 year", time());
+$date = date("Y-m-d", $time);
+
+$db->query(
+    'UPDATE public.member
+	SET data_removal=\'default\'
```
michael-grace, line 31: default?

src/Classes/ServiceAPI/MyRadio_User.php
JP5457: I was going for a future proof approach. That way any services we have that get user user data both inside and outside of myradio are affected.

src/Classes/ServiceAPI/MyRadio_User.php
JP5457: I kind of want to leave it as an option for current members. If someone decides to leave the society half way through a year or has valid reasons for wanting a hidden profile I think it should be an option, however as said here a highly discouraged one.

src/Classes/ServiceAPI/MyRadio_User.php
michael-grace: 👍

src/Classes/ServiceAPI/MyRadio_User.php
michael-grace: 👍

scripts/gdpremail.php
JP5457: This is a change that could definitely be made however i don't really want to put anything into a daemon until its been tested properly and I don't want to test it until it needs to be used because it will email A LOT of people.

scripts/gdprdeleteall.php
JP5457: This reverts users account back to default if they have signed in since the warning email

scripts/gdprdeleteall.php
```diff
@@ -0,0 +1,42 @@
+#!/usr/local/bin/php -q
+<?php
+/**
+ * DO NOT RUN THIS SCRIPT WILLY NILLY
+ * this script will delete the personal data of any members that have been emailed regarding deletion and have not acted
+ * you should probably run the gdpremail script first then wait a week or two
+ * 
+ * run with "php gdprdeleteall.php"
```
LordAro, line 8: perhaps this now needs a slightly less scary name?

scripts/gdprdeleteuser.php
```diff
@@ -0,0 +1,114 @@
...
+            VALUES (1350883, \'deleted\', \'user\', 10, false, \'deleted\')',
+        []
+    );
+}  catch (exception $e) {
+    echo 'deleting user\n';
+}
+$textMemberList = implode(',', $memebersToDelete);
+
+$db->query(
+    'UPDATE schedule.show_credit SET memberid=1350883 WHERE memberid=$1',
```
LordAro, line 44: what is this 1350883 ? Probably should be a named constant?

scripts/gdprdeleteall.php
```diff
@@ -0,0 +1,42 @@
...
+
+use \MyRadio\Database;
+use \MyRadio\MyRadioException;
+use \MyRadio\MyRadioError;
+use \MyRadio\MyRadioEmail;
+use \MyRadio\MyRadio\CoreUtils;
+use \MyRadio\ServiceAPI\MyRadio_User;
+
+ini_set('log_errors', 1);
+ini_set('error_log', '/tmp/php-mailparser-error.log');
```
LordAro, line 19: This seems wrong

scripts/gdprdeleteuser.php
```diff
@@ -0,0 +1,114 @@
...
+
+use \MyRadio\Database;
+use \MyRadio\MyRadioException;
+use \MyRadio\MyRadioError;
+use \MyRadio\MyRadioEmail;
+use \MyRadio\MyRadio\CoreUtils;
+use \MyRadio\ServiceAPI\MyRadio_User;
+
+ini_set('log_errors', 1);
+ini_set('error_log', '/tmp/php-mailparser-error.log');
```
LordAro, line 19: This also seems wrong

scripts/gdprdeleteuser.php
```diff
@@ -0,0 +1,114 @@
...
+    $db->query(
+        'INSERT INTO public.member(
+            memberid, fname, sname, college, receive_email, data_removal)
+            VALUES (1350883, \'deleted\', \'user\', 10, false, \'deleted\')',
+        []
+    );
+}  catch (exception $e) {
+    echo 'deleting user\n';
+}
+$textMemberList = implode(',', $memebersToDelete);
```
LordAro, line 41: this is both not defined, and not used

scripts/gdprdeleteuser.php
```diff
@@ -0,0 +1,114 @@
...
+ini_set('display_errors', 'On');
+
+require_once '../src/Controllers/root_cli.php';
+
+$db = Database::getInstance();
+
+$time = strtotime("-1 year", time());
+$date = date("Y-m-d", $time);
+
+$userid = $argv[1];
```
LordAro, line 29: Given this is the super scary delete script, I'd add a "are you really really sure" confirmation section. Just to help prevent accidents

scripts/gdpremail.php
```diff
@@ -0,0 +1,74 @@
...
+<p>If you are happy for your personal data to be deleted feel free to ignore this eamil.</p>
+--<br/>
+The URY Computing team<br/>
+<br/>
+University Radio York 1350AM 88.3FM<br/>
+---------------------------------------------<br/>
+<a href=“mailto:head.of.computing@ury.org.uk”>head.of.computing@ury.org.uk</a><br/>
+---------------------------------------------<br/>
+On Air | Online | On Tap<br/>
+<a href=“https://ury.org.uk”>ury.org.uk</a>
```
LordAro, line 41: I'd imagine these unicode quote characters won't work so well

scripts/gdpremail.php
```diff
@@ -0,0 +1,74 @@
...
+ini_set('error_log', '/tmp/php-mailparser-error.log');
+ini_set('display_errors', 'On');
+
+require_once '../src/Controllers/root_cli.php';
+
+$db = Database::getInstance();
+
+$warning_email = <<<EOT
+<p>You are getting this email because you have not logged into MyRadio in over a year</p>
+<p>In one week all of your personally identifiable data that is not required for webstudio or our public facing websites to function as advertsied will be deleted.</p>
```
LordAro, line 30: >advertsied

scripts/gdpremail.php
```diff
@@ -0,0 +1,74 @@
...
+<a href=“mailto:head.of.computing@ury.org.uk”>head.of.computing@ury.org.uk</a><br/>
+---------------------------------------------<br/>
+On Air | Online | On Tap<br/>
+<a href=“https://ury.org.uk”>ury.org.uk</a>
+EOT;
+
+$time = strtotime("-1 year", time());
+$date = date("Y-m-d", $time);
+
+$memebersToEmail = $db->fetchAll(
```
LordAro, line 47: >$memebers

(several other places too)

src/Classes/ServiceAPI/MyRadio_User.php
```diff
@@ -785,6 +825,16 @@ public function getRequirePasswordChange()
         return $this->require_password_change;
     }
 
+    public function getSignedGDPR()
+    {
+        return $this->gdpr_accepted;
+    }
+
+    public function getHideProfile()
```
LordAro, line 833: Feels like a poor name for a function that returns a bool. `isHidden` ?

Similarly for `getSignedGDPR`

scripts/gdprdeleteall.php
```diff
@@ -0,0 +1,42 @@
...
+ini_set('display_errors', 'On');
+
+require_once '../src/Controllers/root_cli.php';
+
+$db = Database::getInstance();
+
+$time = strtotime("-1 year", time());
+$date = date("Y-m-d", $time);
+
+$db->query(
```
LordAro, line 29: Still pretty confident this should be using MyRadio set functions, rather than direct database bashing for the majority of cases.

LordAro (commented):
Yeah, this feels like a much more reasonable approach to me, thanks for taking the time to address our concerns :)

Do update the PR description to keep it up to date, and don't forget to update previous review comments (mark them as resolved when.. they are) so you don't get overwhelmed by them

Also, codestyle :)

scripts/gdprdeleteuser.php
JP5457: It has to be an int for database reasons. 1350883 is just a futureproofed number made by smashing together our fm and am frequencies.

scripts/gdpremail.php
JP5457: oops

scripts/gdprdeleteall.php
JP5457: personally I'm against that. I think we'd be adding quite a lot of code to our classes which will make our codebase even more confusing without it having much benefit over the database queries.

scripts/gdprdeleteall.php
JP5457: I don't mind scaring people off if you don't know what you're doing you shouldn't be running the script

scripts/gdprdeleteall.php
LordAro: I'm not talking about adding specific functions for this usecase, but using the ones that are already there - setCollege, pulling out a list of shows a particular member has credits on and updating all of those, etc etc

scripts/gdprdeleteuser.php
ColinRoitt: change made

scripts/gdprdeleteall.php
ColinRoitt: Future to do

scripts/gdprdeleteuser.php
ColinRoitt: changes made

src/Controllers/gdprdelete.php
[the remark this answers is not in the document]
ColinRoitt: Can be refactored, future todo.

ColinRoitt:
Future non-essential changes moved to https://github.com/UniversityRadioYork/MyRadio/issues/1137

PR #1138, opened 2023-11-15
Update MyRadio_Scheduler.php - Fixes the 10 week limit for semesters

hitime1234:
updates 10 week limit for the scheduler for 11 weeks 
To work for semesters which are 11 weeks
https://www.york.ac.uk/students/studying/semesters/#:~:text=You%20will%20be%20expected%20to,whilst%20you%20are%20undertaking%20assessments.

Nya :3
- Ren

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Scheduler.php
+++ b/src/Classes/ServiceAPI/MyRadio_Scheduler.php
@@ -188,10 +188,6 @@ class MyRadio_Scheduler extends ServiceAPI
         $term = new MyRadio_Term($term_id);
         $start_day = $term->getTermStartDate() + ($time['day'] * 86400);
         //Iterate over each week
-        // Changes it to get number of weeks in a semester from Myradio_term
-
-        
-
         for ($i = 1; $i <= $term->getTermWeeks(); ++$i) {
             $day_start = $start_day + (($i - 1) * 7 * 86400);
 
--- a/src/Classes/ServiceAPI/MyRadio_Scheduler.php
+++ b/src/Classes/ServiceAPI/MyRadio_Scheduler.php
@@ -187,8 +187,10 @@ class MyRadio_Scheduler extends ServiceAPI
         $conflicts = [];
         $start_day = (new MyRadio_Term($term_id))->getTermStartDate() + ($time['day'] * 86400);
         //Iterate over each week
-        //Changes to 11 weeks for semesters
-        for ($i = 1; $i <= 11; ++$i) {
+        // Changes it to get number of weeks in a semester from Myradio_term
+        $num_weeks = getTermWeeks()
+
+        for ($i = 1; $i <= $num_weeks; ++$i) {
             $day_start = $start_day + (($i - 1) * 7 * 86400);
 
             //Get the start time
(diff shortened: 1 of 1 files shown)


src/Classes/ServiceAPI/MyRadio_Scheduler.php
```diff
@@ -187,7 +187,8 @@ public static function getScheduleConflicts($term_id, $time)
         $conflicts = [];
         $start_day = (new MyRadio_Term($term_id))->getTermStartDate() + ($time['day'] * 86400);
         //Iterate over each week
-        for ($i = 1; $i <= 10; ++$i) {
+        //Changes to 11 weeks for semesters
+        for ($i = 1; $i <= 11; ++$i) {
```
markspolakovs, line 191: So this works, but there's no guarantee that semesters will always be 11 weeks - is there a way that we can avoid hard-coding a number here completely?

hitime1234: I changed it to use the num_weeks from Myradio_term. 
Is this okay or do I need to get the number weeks from somewhere else

markspolakovs: You're on the right track. As written it won't work, because`getTermWeeks` is a method on `MyRadio_Term` so you need to call it on a term object (`$term->getTermWeeks()`). You'd need to do something like `(new MyRadio_Term($term_id))->getTermWeeks()`.

However, since there's already another MyRadio_Term constructor above, you could remove the duplication:

```php
$term = new MyRadio_Term($term_id);
$start_day = $term->getTermStartDate() + ($time['day'] * 86400);

for ($i = 1; $i <= $term->getTermWeeks(); ++$i) {
// ...
}
```

src/Classes/ServiceAPI/MyRadio_Scheduler.php
```diff
@@ -187,7 +187,10 @@ public static function getScheduleConflicts($term_id, $time)
         $conflicts = [];
         $start_day = (new MyRadio_Term($term_id))->getTermStartDate() + ($time['day'] * 86400);
         //Iterate over each week
-        for ($i = 1; $i <= 10; ++$i) {
+        // Changes it to get number of weeks in a semester from Myradio_term
+        $num_weeks = getTermWeeks()
```
JP5457, line 191: probably needs a ;

hitime1234: oopise daisy

should be fixed now

src/Classes/ServiceAPI/MyRadio_Scheduler.php
```diff
@@ -185,9 +185,14 @@ public static function getScheduleConflicts($term_id, $time)
...
         self::initDB();
         $conflicts = [];
-        $start_day = (new MyRadio_Term($term_id))->getTermStartDate() + ($time['day'] * 86400);
+        $term = new MyRadio_Term($term_id);
+        $start_day = $term->getTermStartDate() + ($time['day'] * 86400);
         //Iterate over each week
-        for ($i = 1; $i <= 10; ++$i) {
+        // Changes it to get number of weeks in a semester from Myradio_term
+
+
```
markspolakovs, line 193: This comment isn't necessary - once this PR is merged, the change will have been applied, so there's no need for it (the commit message documents it enough that you don't need a comment).

Otherwise looks good to me, thank you!

markspolakovs (approved):

PR #937, opened 2020-09-24
Sign-in changes for contact tracing

markspolakovs:
* Sign-ins now have a location
* Guests can be signed in (with a freeform text box)

Next step will be to make a UI for getting at the data, but this is a good enough start for a review.

PR #1140, opened 2023-12-11
Remove location option from sign-ins

markspolakovs:
Partly reverts #937 - removes references to "track & trace" and removes the location dropdown, but keeps the guests entry button as that may still be useful.


michael-grace:
Didn't we decide to keep the dropdown tho

markspolakovs:
Closing in favour of #1145

PR #1142, opened 2024-01-02
Upgrade development Dockerfile to PHP 8 + setup fixes

markspolakovs:
Two changes in this PR:
* First upgrades the Dockerfile to PHP 8.2. I had to remove a bunch of dev-only Composer dependencies that don't support PHP 8 (and some don't exist at all anymore) - the only risky one is codeception which is a major version bump 
* Second onwards is all the changes I had to make to the setup flow to get it to succeed in one go - mainly making all the database operations idempotent and a few other smaller fixes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM php:7.4-apache
+FROM php:8.1-apache
 
 RUN apt-get update && apt-get install -y libpq-dev libpng-dev libjpeg-dev libldap-dev unzip \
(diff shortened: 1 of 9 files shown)


LordAro (commented):
PHP 8.1 is already out of (active) support, why not 8.2 or 8.3 ?

https://endoflife.date/php

markspolakovs:
🤦 I meant 8.2, not sure why I wrote 8.1...

ashhhleyyy (approved):

PR #1146, opened 2024-02-17
fun fact: this has been running in prod for a while

michael-grace:
few things:

- few updates for running in the container after thunderhorn said "no more myradio for you"
- updates for the off air studio booking app, such as myradio issuing jwts to do better authentication than what we had

Then, we can have what's running be up to date with github again, and can then give it to jenkins now the new version is running _mostly_ ok.

### DIFF
--- a/Dockerfile.prod
+++ b/Dockerfile.prod
@@ -1,4 +1,4 @@
-FROM php:8.2-apache
+FROM php:7.4-apache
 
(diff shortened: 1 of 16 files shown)


Dockerfile.prod
```diff
@@ -0,0 +1,55 @@
+FROM php:7.4-apache
```
ashhhleyyy, line 1: this will need updating to 8.2 when #1142 is merged

ashhhleyyy (approved):

PR #1149, opened 2024-05-20
It doesn't actually work on PHP 8

michael-grace:
So, here's reverting it for now. If you do intend to fix it, please use myradio-staging (you can build a specific branch using jenkins) and _make sure_ it builds on there before pushing it to master where it'll build for production.

you also need to check the api, check /api or /api-staging as appropriate to check it doesn't break that (currently it does)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM php:8.2-apache
+FROM php:7.4-apache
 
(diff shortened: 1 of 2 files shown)


markspolakovs:
> It doesn't actually work on PHP 8

What specifically doesn't work about it?

PR #1161, opened 2024-09-05
terms api

michael-grace:
lets the off air studio booking calendar import the week names from myradio, very nice


ashhhleyyy (approved):
works for me :+1:

PR #1165, opened 2024-09-09
Fix graphql api

ashhhleyyy:
Updates php-graphql (not necessary for the fix, but still good to have)
Fixes the TypeError on graphql queries

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
         "soundasleep/html2text": "~0.5",
         "twig/twig": "~2.4",
         "ext-xmlwriter": "*",
-        "webonyx/graphql-php": "^0.13.8",
+        "webonyx/graphql-php": "^15.0.0",
         "ext-json": "*",
         "spatie/icalendar-generator": "^2.1",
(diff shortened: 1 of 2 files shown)


markspolakovs (commented):
composer.lock?

ashhhleyyy:
the repo doesn't have one?

markspolakovs (approved):
Indeed it doesn't, whoops. (Maybe it should, but that's a separate issue.)

LGTM, though I am slightly concerned by the jump from 0.13.8 to 15.0.0 (it's not as big as it looks - seemingly upstream went from 0.13 to 14.0 - but still a major bump). Have you had a chance to test the new version? 

Approving on the basis that you've tested it.

ashhhleyyy:
I've tested with the currently failing queries from radioplayer and icetracks2, along with playing around in graphiql with other queries, and I didn't find anything to be broken - I've also had a skim-read through the changelog and for the two breaking releases (14.0.0 and 15.0.0) and didn't see anything that immediately stood out as being an issue

PR #1166, opened 2024-09-09
Support building the docker images with podman

ashhhleyyy:


markspolakovs (commented):
Looks good, though I'm a bit unsure why it's necessary - does podman not handle multi-line RUN instructions?

ashhhleyyy:
The `service apache2 restart` command is unnecessary and causes apache to try and start up during container build, which fails to bind to port 80 (I assume because of podman rootless builds)

markspolakovs (approved):

PR #1170, opened 2024-09-15
fix: fix user signup for reals this time

ashhhleyyy:


markspolakovs (approved):
LGTM assuming `nname` has a sane default in the database

hitime1234:
> LGTM assuming `nname` has a sane default in the database

nname is null default

PR #1173, opened 2024-09-16
Revert nicknames until tested further

ashhhleyyy:
It has caused many issues due to not being fully tested, and we need MyRadio to be completely functional for freshers.

### DIFF
--- a/src/Templates/MyRadio/getOnAir.twig
+++ b/src/Templates/MyRadio/getOnAir.twig
@@ -9,7 +9,7 @@
 
 <!--Pay Membership-->
 <li>
-  <a href="{{ config.payment_url }}" target="_blank" title="Before we can let you use our shiny and expensive studio, you'll need to pay membership. Click here to do so on the YORKSU website now.">Become a Paid Member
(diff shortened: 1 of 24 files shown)


src/Templates/MyRadio/getOnAir.twig
```diff
@@ -9,7 +9,7 @@
 
 <!--Pay Membership-->
 <li>
-  <a href="{{ config.payment_url }}" target="_blank" title="Before we can let you use our shiny and expensive studio, you'll need to pay membership. Click here to do so on the YORKSU website now.">Become a Paid Member
+  <a href="{{ config.payment_url }}" target="_blank" title="Before we can let you use our shiny and expensive studio, you'll need to pay membership. Click here to do so on the YUSU website now.">Become a Paid Member
```
markspolakovs, line 12: This change can stay, although we probably want to use the official capitalisation
```suggestion
  <a href="{{ config.payment_url }}" target="_blank" title="Before we can let you use our shiny and expensive studio, you'll need to pay membership. Click here to do so on the York SU website now.">Become a Paid Member
```

markspolakovs (approved):

src/Templates/MyRadio/getOnAir.twig
hitime1234: This was in the pending pr for nickname

PR #1174, opened 2024-09-16
YUSU -> York SU

markspolakovs:


ashhhleyyy (approved):
it will always be yusu in my heart :heart:

PR #1177, opened 2024-09-16
fix: workaround broken email sending

ashhhleyyy:
This should fix email formatting until we get round to rewriting email sending .

Tested locally with mailhog and the broken formatting doesn't show up anymore


ashhhleyyy:
Superseded by #1178

PR #1189, opened 2024-10-02
Split the I want to section

AlexNimzy:
I added a section called Tools + Merch because I didn't know what section the latter went in. I'm free to change it.

### DIFF
--- a/src/Menus/menu.json
+++ b/src/Menus/menu.json
@@ -52,51 +52,6 @@
                     ]
                 },
                 {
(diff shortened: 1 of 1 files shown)


JP5457:
Looks really good! Could we move the three sections to all be in different columns? maybe contacts under my station and requests under my services?

ColinRoitt (changes requested):

JP5457 (approved):

PR #1190, opened 2024-10-02
events can have a different owner

alyxbb:
fun fact: myradio only makes me want to scream 99% of the time

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Event.php
+++ b/src/Classes/ServiceAPI/MyRadio_Event.php
@@ -136,6 +136,21 @@ class MyRadio_Event extends ServiceAPI
     }
 
     /**
+     * @param int
+     */
+    public function setHostId(int $hostId): void
+    {
+      self::$db->query('UPDATE public.events
+          SET hostid = $2
+          WHERE eventid = $1', [
+          $this->getID(),
+          $hostId
+      ]);
+      $this->hostId = $hostId;
+      $this->updateCacheObject();
+    }
+
+    /**
      * @return string
      */
     public function getDescriptionHtml(): string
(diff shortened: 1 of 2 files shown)


ashhhleyyy:
skill issue

src/Classes/ServiceAPI/MyRadio_Event.php
```diff
@@ -277,15 +295,15 @@ public static function create($data = [])
             }
         }
 
-        $hostid = MyRadio_User::getCurrentOrSystemUser()->getID();
+        //$hostid = MyRadio_User::getCurrentOrSystemUser()->getID();
```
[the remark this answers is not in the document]
alyxbb: oops, meant to delet it

markspolakovs:
Also, I know that this is a student society and not a workplace, but could we please aim for a slightly higher level of professionalism when it comes to commit messages. The messages you write are read by other people, which may include (a) your future employers and (b) the people who wrote the code that "makes you scream" (in the case of the files this PR modifies, that would be me). There's more appropriate places to vent about the quality of the code you're maintaining than a public PR description.

src/Classes/ServiceAPI/MyRadio_Event.php
```diff
@@ -135,6 +135,21 @@ public function setEndTime(int $endTime): void
...
+     * @param int
+     */
+    public function setHostId(int $hostId): void
+    {
+      self::$db->query('UPDATE public.events
+          SET hostid = $2
+          WHERE eventid = $1', [
+          $this->getID(),
+          $hostId
+      ]);
```
markspolakovs, line 148: Indentation seems a bit off here

markspolakovs (approved):

src/Classes/ServiceAPI/MyRadio_Event.php
alyxbb: its the same as everywhere else

PR #1191, opened 2024-10-04
[WIP] Check for concurrent edits of showplans

markspolakovs:
Add a simple mechanism to check if two people are editing a showplan at the same time, by storing the time it was last modified.

Will go together with a WebStudio change that will, when opened, get the last modified time from the server, and then send it back on every updateShowPlan call. The server will only accept the change if the time still matches (guarded by a `SELECT FOR UPDATE`), otherwise WebStudio will force the user to reload.

This can be a bit annoying, but the alternative is showplan corruption. Perhaps we can make WebStudio periodically poll for new changes to make it less annoying.

### DIFF
--- a/schema/patches/19.sql
+++ b/schema/patches/19.sql
@@ -1,0 +1,9 @@
+BEGIN;
+ALTER TABLE schedule.show_season_timeslot
+  ADD COLUMN showplan_last_modified TIMESTAMPTZ DEFAULT NULL;
+
+UPDATE myradio.schema
+SET value = 14
+WHERE attr='version';
+
+COMMIT;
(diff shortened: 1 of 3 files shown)


schema/patches/19.sql
```diff
@@ -0,0 +1,9 @@
+BEGIN;
+ALTER TABLE schedule.show_season_timeslot
+  ADD COLUMN showplan_last_modified TIMESTAMPTZ DEFAULT NULL;
+
+UPDATE myradio.schema
+SET value = 14
```
markspolakovs, line 6:
```suggestion
SET value = 19
```

wat

PR #1194, opened 2024-10-17
Show booking availability in training form

ColinRoitt:
Replaces the schedule preview with the booking form from the booking microservice site. This URL is set in the `config.php`.

Resolves issue with https://github.com/orgs/UniversityRadioYork/projects/10/views/6?pane=issue&itemId=80365237&issue=UniversityRadioYork%7CMyRadio%7C1185  but does not close ticket


ashhhleyyy (approved):
works on my machine

PR #1201, opened 2024-12-06
improve search

alyxbb:
search can now take the search query as a query paramter, enabling search for special characters

### DIFF
--- a/src/Classes/ServiceAPI/MyRadio_Podcast.php
+++ b/src/Classes/ServiceAPI/MyRadio_Podcast.php
@@ -903,18 +903,32 @@ class MyRadio_Podcast extends MyRadio_Metadata_Common
 
     /**
      * Searches searchable *text* metadata for the specified value. Does not work for image metadata.
-     *
+     * if $q is set, then $path_query must be set to "search"
+
+     * 
      * @todo effective_from/to not yet implemented
      *
-     * @param string $query          The query value.
+     * @param string $path_query     The query value encoded in the path (DEPRECATED).
+     * @param string $q              The query value as a query string. to use this, $path_query must be set to "search"
      * @param array  $string_keys    The metadata keys to search
      * @param int    $effective_from UTC Time to search from.
      * @param int    $effective_to   UTC Time to search to.
      *
      * @return array The shows that match the search terms
      */
-    public static function searchMeta($query, $string_keys = null, $effective_from = null, $effective_to = null)
+    public static function searchMeta($path_query, $q="", $string_keys = null, $effective_from = null, $effective_to = null)
     {
+        if($path_query != "search" && $q != "") {
+            throw new MyRadioException(
+                "the path_query must be set to 'search' if q is set in the query string",
+                400
+            );
+        }
+        if ($path_query == "search" && $q != ""){
+            $query = $q;
+        } else {
+            $query = $path_query;
+        }
         if (is_null($string_keys)) {
             $string_keys = ['title', 'description', 'tag'];
         }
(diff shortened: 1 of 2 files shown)


src/Classes/ServiceAPI/MyRadio_Show.php
```diff
@@ -1028,18 +1028,31 @@ public static function getMostListened($date = 0)
 
     /**
      * Searches searchable *text* metadata for the specified value. Does not work for image metadata.
+     * if $q is set, then $path_query must be set to "search"
      *
      * @todo effective_from/to not yet implemented
      *
-     * @param string $query          The query value.
+     * @param string $path_query     The query value encoded in the path (DEPRECATED).
+     * @param string $q              The query value as a query string. to use this, $path_query must be set to "search"
```
JP5457, line 1036: you've included this comment twice however you haven't explained why this behaviour exists

src/Classes/ServiceAPI/MyRadio_Show.php
```diff
@@ -1028,18 +1028,31 @@ public static function getMostListened($date = 0)
...
+            throw new MyRadioException(
+                "the path_query must be set to 'search' if q is set in the query string",
+                400
+            );
+        }
+        if ($path_query == "search" && $q != ""){
+            $query = $q;
+        } else {
+            $query = $path_query;
+        }
```
JP5457, line 1055: what is the point of adding a parameter if in both cases $query gets set to one or the other?

src/Classes/ServiceAPI/MyRadio_Podcast.php
```diff
@@ -903,18 +903,32 @@ public function getCover()
...
+            throw new MyRadioException(
+                "the path_query must be set to 'search' if q is set in the query string",
+                400
+            );
+        }
+        if ($path_query == "search" && $q != ""){
+            $query = $q;
+        } else {
+            $query = $path_query;
+        }
```
JP5457, line 931: see my comment on myRadio_Show

src/Classes/ServiceAPI/MyRadio_Podcast.php
```diff
@@ -903,18 +903,32 @@ public function getCover()
...
      * Searches searchable *text* metadata for the specified value. Does not work for image metadata.
-     *
+     * if $q is set, then $path_query must be set to "search"
+
+     * 
      * @todo effective_from/to not yet implemented
      *
-     * @param string $query          The query value.
+     * @param string $path_query     The query value encoded in the path (DEPRECATED).
+     * @param string $q              The query value as a query string. to use this, $path_query must be set to "search"
```
JP5457, line 912: see my comment on myRadio_Show

JP5457 (changes requested):
You've added multiple redundant comments that do nothing to clarify how the function works and you have added a new parameter for seemingly no reason.

src/Classes/ServiceAPI/MyRadio_Show.php
alyxbb: this is because of myradio weirdness. in the service api when called the first arg is gotten from the path section of the url. subsequent args are gotten from query parameters. to get a query parameter you need a second arg. we set query based on whether they are using query parameters or path to maintain backwards compatibility

src/Classes/ServiceAPI/MyRadio_Show.php
alyxbb: how parameters work in the serviceapi is weird. its a bit clearer in the autogenerated api docs but i will add more details to it

src/Classes/ServiceAPI/MyRadio_Show.php
```diff
@@ -1028,18 +1028,31 @@ public static function getMostListened($date = 0)
 
     /**
      * Searches searchable *text* metadata for the specified value. Does not work for image metadata.
+     * if $q is set, then $path_query must be set to "search"
```
JP5457, line 1031:
```suggestion
     * if $q is set, then $path_query must be set to "search". This allows query to be given in path or in parameters.
```

JP5457 (approved):
I'm happy with the code although I'd appreciate it if you included the details about the path in a comment in the code base before merging. This is to avoid confusion in future as its very unclear why this code exists without that knowledge.

PR #1209, opened 2025-04-16
Add patch for cancel timeslot coming through API

ColinRoitt:
I genuinely have no idea if this is even remotely secure. Please tell me (and then the right way to do it).


ashhhleyyy (commented):
i think this is good, maybe just change the user check to `MyRadio_User::getCurrentUser() == null` just to avoid hardcoding the session parameter name

ColinRoitt:
Fair comment but this is already done all over both this file and other service API files so its kind of the pattern it seems.

ColinRoitt:
actually u were right I did it the right way - all tested and functional

JP5457 (approved):
lgtm

PR #1220, opened 2025-06-10
add pgadmin container to development docker-compose

ashhhleyyy:
does what it says on the tin


hitime1234 (approved):
The docker compose looks correct,
We might want to mention this readme tho that's completely a suggestion on my part

PR #1232, opened 2026-02-08
fix(docker): add --no-security-blocking

ashhhleyyy:
makes the CI and everything work again because updating twig would be a major version bump


archessmn (approved):
LGBTQ

PR #1235, opened 2026-05-18
chore: bump packages

ashhhleyyy:
bumping as many packages as we can


Dockerfile.prod
LooFifteen: I feel like we can't merge until this flag is sorted.

hitime1234: Tell if I am wrong but this because of twig 2

